PR #8, opened 2020-02-23
feat(moderation): per-user slowmode CC

JordiHakimZiyech:
slowmode per user command
hopefully i've PR'ed it correctly

### DIFF
--- a/maintenance/slowmode.cc.lua
+++ b/maintenance/slowmode.cc.lua
@@ -1,0 +1,12 @@
+{{/* This command will set a slowmode for a specific User
+ Trigger Type: Regex -- Trigger: .*   */}}
+ 
+{{/* change the number below to the amount of seconds that the slowmode should be */}}
+{{ $time := 10 }} 
+{{ if eq (toInt (dbGet 2 "slowmode").Value) 0 }}
+{{ dbSetExpire 2 "slowmode" 1 $time }}
+{{ else }}
+{{ if eq (toInt (dbGet 2 "slowmode").Value) 1 }} 
+{{ deleteTrigger 0 }}
+{{ end }}
+{{ end }}


Jo3-L:
Hey, thanks for the PR. Some comments though:
1. How is this different from Discord's slowmode / similar implementation in automod(v2)?
2. Please follow the guidelines laid out [here](https://github.com/Jo3-L/yagpdb-cc/blob/master/CONTRIBUTING.md).

Satty9361:
I think this is a per user thing while discord slowmode is for all roles (execpt manage messages etc). In that way it is quite different. Much different implementation as well because this deletes messages while actual slowmode prevents sending them altogether.

JordiHakimZiyech:
yeah it is per user. And yes, it deletes the message. It is possible to prevent them sending messages, but that requires giving a role, and some people just don't like even more roles.

Satty9361:
You need to just format it according to this i guess @JordiHakimZiyech 
Link : https://github.com/Jo3-L/yagpdb-cc/blob/master/CONTRIBUTING.md#leading-comment-syntax

Jo3-L:
In addition to what Satty said, you should put this in a new folder, maybe `moderation` or something. Concept / usage is not suitable for any of the current folders other than `util`, and that's a stretch as well.

Also, looking at the code, 
```lua
{{ $time := 10 }} 
{{ if eq (toInt (dbGet 2 "slowmode").Value) 0 }}
{{ dbSetExpire 2 "slowmode" 1 $time }}
{{ else }}
{{ if eq (toInt (dbGet 2 "slowmode").Value) 1 }} 
{{ deleteTrigger 0 }}
{{ end }}
{{ end }}
```
Does not seem to be user-based (though that could be changed). I'll be leaving this unmerged for now.

JordiHakimZiyech:
yeah. My bad. i'll change it in a bit and i'll drop it in a new file then

Jo3-L:
Closing due to inactivity, please feel free to make a new PR if you wish.

PR #9, opened 2020-03-02
Update emoji.cc.lua

alikuxac:
The emoji command I've made


Jo3-L:
Thanks for the PR, but this is already available (more or less same thing) in the `bigemoji` CC, located at [util/bigemoji](https://github.com/Jo3-L/yagpdb-cc/blob/master/util/bigemoji.cc.lua).

PR #31, opened 2020-06-12
Create BirthdayCC.Lua

### LINKED ISSUE #30: Birthday system
## Birthday System
A birthday system has often been requested in the support server, I feel that this repo is an appropriate place to put it.

## Description
A user may set their birthday using a one-use command (to prevent abuse). Every day, the bot checks for birthdays on that date. If there are, a notification is sent in a given channel and an optional role is given to all users with birthdays on that date (removed during the next day's check).

Admins may force set birthdays if a user has set their birthday to the wrong date.

## Implementation
**Command for users to set their birthday**
First, we check whether this user has a birthday set. Next, the date they have given is validated. If it is valid, we add their user ID to a database entry with the key being in the format `DD-MM-YYYY`.

**Interval trigger**
Firstly, we get the DB entry `latestBirthdays`. We loop over all user IDs in this entry, and remove the birthday role from them. Then, we get the current date, then retrieve the entry for that day. We update the DB entry `latestBirthdays` with the value of the entry just retrieved. After, we loop over the user IDs for today, sending a single congratulation message and then adding the birthday role to all of them.

We may need to ensure that we don't hit ratelimits during this process.

Pedro-Pessoa:
Closes #30


Jo3-L:
Nit: file names should be `camelCased` and there should be a `.` in front of CC - `birthday.cc.lua`.

Pedro-Pessoa:
Done :)

PR #35, opened 2020-06-12
Lock down cc

Pedro-Pessoa:


util/lockdown.cc.lua
```diff
@@ -0,0 +1,103 @@
+{{/* Lockdown CC
```
Jo3-L, line 1: Newline after `{{/*` please

PR #36, opened 2020-06-13
Lockdown CC

Pedro-Pessoa:


moderation/lockdown.cc.lua
```diff
@@ -0,0 +1,104 @@
+{{/*
+Lockdown CC
```
Jo3-L, line 2: Indent here, otherwise looks fine to me

Pedro-Pessoa:
Done :)

PR #43, opened 2020-07-09
Birthday Command improvement

Pedro-Pessoa:
Added a -delbday command
Created a readme because of 10k limit
Fixed two bugs related to underage users


Jo3-L:
Looks good!

PR #67, opened 2020-08-15
Update guessTheNumber.cc.lua

sbusse61:
Moved {{ dbDel 0 "NR" }} to after the embed message.  Database value was being deleted prior to message being created, and was always returning 0.


DaviiiD1337:
Ill pr the v2 no need to merge joe

Jo3-L:
Superseded by #69.

sbusse61:
Thank you both. Didn't know v2 was in the works.

PR #73, opened 2020-08-31
Update README.md

iratekalypso:
(?i) is breaking the CC it seems.

### DIFF
--- a/suggestion/v2/README.md
+++ b/suggestion/v2/README.md
@@ -4,11 +4,11 @@
 
 ### Trigger -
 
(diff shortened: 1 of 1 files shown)


Jo3-L:
context? `(?i)` is fine here syntactically speaking, don't see any reason why it'd break the CC - need an explanation before I can merge.

iratekalypso:
After troubleshooting in #priority-support on the YAGPDB Discord Server, that was the one thing we needed to change to make the CC work. It seemed to break the REGEX for some reason while testing and setting it up. Once `(?i)` was removed the CC functioned fine.

Jo3-L:
I dunno, `(?i)` should not be breaking in any way there - I suspect you probably just caught Yag during a time of high latency but I'll wait for Satty before closing this.

Satty9361:
[https://regex101.com/r/yWcrlO/1](Regex101 Link)

As you can see the regex is correct. (?i) allows the regex to be case insensitive. 

That must have been definitely some other issue. If you can reproduce any issue, feel free to tag me here/in discord with an invite. 

Will close for now.

PR #83, opened 2020-10-06
Role-based commands for extending rolemenu

theo-brown:
I use these two commands in my server to assign people extra roles when choosing roles from a rolemenu, and to notify channels when users receive new roles.

They receive a lot of use in my channel and I think others might find them useful too!


Jo3-L:
just a tip - these should not extend rolemenu, rather, they should replace the rolemenu functionality completely as it may cause weird edge cases (i.e rolemenu is on cooldown, doesn't give one role but CC doesn't have this issue and gives the other one).

personally, I feel this is a bit too specific in its current form, perhaps allow an arbitrary amount of roles to be assigned / removed ex.
```go
{{ $rolesToChange := cslice 1 2 3 }}
{{ $messageId := 123 }}
{{/* code to facilitate a single reaction adding / removing all these roles */}}
```

Jo3-L:
Going to close for now @theo-brown, if you're interested in implementing the changes I specified above please re-open.

PR #88, opened 2020-10-09
Change trigger for afk.cc.lua

NaruDevnote:
Didn't think I would be doing this since its only one character but yeah :))


Jo3-L:
lgtm, ty

PR #90, opened 2020-10-14
Restructure and reformat commands and READMEs

thewilloftheshadow:
Mass draft PR for the changes I’m doing as they’re completed for feedback and noting changes :)

Closes #40
Fixes #53
Fixes #49
Fixes #91

PR #93, opened 2020-10-27
Command descriptions

Crenshaw7860:
I started a description thingy, I think this will be helpful try and keep the exeCC check box when you write yours.


Jo3-L:
not gonna merge this due to #90 as descriptions will be in their own file. thanks for the PR though!

PR #105, opened 2020-11-25
Message link preview V2(ish)

10by10pixel:
It's just the quote command from here: https://discord.com/channels/166207328570441728/551526991250325544/576765531974008832 but with the regex fixed. If someone's already done this, I couldn't find it.


Jo3-L:
Thanks for the PR, but I'm not going to merge this as Crafter has a far better message link CC up on his GitHub which is also what we use in the Yag server. Also - we do currently have a message link CC here, but I'm thinking of removing that and just linking to Crafter's.

PR #106, opened 2020-11-27
Update regex in some commands

alikuxac:
Some commands have old regex. So I decided to modified these.
All new regex are tested and it works perfectly.


Jo3-L:
Thanks for the PR!

PR #123, opened 2020-12-14
Bump CC

WickedWizard3588:

This is a Bump CC for people who want to grow their Server. These codes are not standalone and need to be used together. I have specified everything as per the [Contributing.md](../CONTRIBUTING.md).

Features:- 
1. Pings the role when its the time to bump

2.  Easy to use and customize

3. A command that shows the Rank and LeaderBoard.

4. The Reaction CC, which helps to add and remove Bumping Roles.

5. An Interval CC that sends messages when the time has come to bump. (Mentioned above) Also, it strikes out the message in the channel after the bumping has been done.

### Code Status:-
The Codes are given in a 3 separate files, which are:-
1. **Reaction CC:-** Adds and Removes the roles when bumping is successful

2. **An Interval CC:-** Edits the Voice Channel name, along with sending the Bump Message

3. **Regex CC:-** This contains the code of `!d lb`, `!d bump` and `!d rank`. This also, strikes out the message sent by the Interval CC.

This is not the end of the list, and I'm planning to add more functionality.

Thanks

**Status**

- [✔] Code changes have been tested on an instance of YAGPDB, or there are no code changes
- [ ✔] I have read and followed the [Contribution guide](../CONTRIBUTING.md)


JordiHakimZiyech:
small notes:
im not so sure _if_ custom commands should be published here that use other bots, but thats my own opinion
2nd point is that apparantly its against disboard's guidelines so that might be an issue

Jo3-L:
Thanks for the PR, but I'm not going to accept this here as Disboard bump CCs have often been very finnicky (cooldowns are impossible to get correct without triggering on the Disboard bot itself) and prone to breakage. Also, as Jordi said above, Disboard has been very unclear about whether bump leaderboards are against its guidelines or not.

PR #126, opened 2020-12-15
Create giveRoleAfterDelay

JordiHakimZiyech:
gives role after delay cuz why not huh

### DIFF
--- a/util/giveRoleAfterDelay
+++ b/util/giveRoleAfterDelay
@@ -1,0 +1,37 @@
+{{/*
+	This command gives a role *after* delay
+	Usage: "-timedrole @user [time] role_name" [time] can be 10d, 10m, 10s etc.
+ 
+	Recommended trigger: timedrole
+	Trigger type: Command
+ 
+	
+*/}}
+{{/*ACTUAL CODE, DO NOT TOUCH*/}}
+{{ if not .ExecData }}
+{{ $args := parseArgs 3 "Wrong usage: `-timedrole (@user) (time) (role name)`" 
+ (carg "userid" "member to give role" )
+ (carg "duration" "time till role" )
+ (carg "string" "role name" ) }}
+{{ $search := slice .CmdArgs 2 | joinStr "" | lower }} 
+ {{ $role := 0 }} {{/* Role variable */}}
+ {{/* Search for roles */}}
+ {{- range .Guild.Roles -}}
+ {{ if not $role }}
+ {{ $rName := lower .Name }}
+ {{ if or (eq $search $rName) (in $rName $search) }}
+ {{ $role = . }}
+ {{ end }}
+ {{ end }}
+ {{- end -}}
+{{ if $role }}
+{{ $a := (toDuration ($args.Get 1)).Seconds }}
+{{ if not (hasRoleName ($args.Get 2)) }}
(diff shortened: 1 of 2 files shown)


util/giveRoleAfterDelay
```diff
@@ -0,0 +1,37 @@
...
+ 
+	Recommended trigger: timedrole
+	Trigger type: Command
+ 
+	
+*/}}
+{{/*ACTUAL CODE, DO NOT TOUCH*/}}
+{{ if not .ExecData }}
+{{ $args := parseArgs 3 "Wrong usage: `-timedrole (@user) (time) (role name)`" 
+ (carg "userid" "member to give role" )
```
Jo3-L, line 13: `member` carg and access `User` field. `userid` doesn't validate.

util/giveRoleAfterDelay
```diff
@@ -0,0 +1,37 @@
...
+ (carg "duration" "time till role" )
+ (carg "string" "role name" ) }}
+{{ $search := slice .CmdArgs 2 | joinStr "" | lower }}
```
Jo3-L, line 16: choose one and be consistent, `parseArgs` or manually parsing

util/giveRoleAfterDelay
```diff
@@ -0,0 +1,37 @@
...
+ {{- range .Guild.Roles -}}
+ {{ if not $role }}
+ {{ $rName := lower .Name }}
+ {{ if or (eq $search $rName) (in $rName $search) }}
+ {{ $role = . }}
+ {{ end }}
+ {{ end }}
+ {{- end -}}
+{{ if $role }}
+{{ $a := (toDuration ($args.Get 1)).Seconds }}
```
Jo3-L, line 28: no need, already a duration.

util/giveRoleAfterDelay
```diff
@@ -0,0 +1,37 @@
...
+{{ if not (hasRoleName ($args.Get 2)) }}
+{{ execCC .CCID nil $a (sdict "user" ($args.Get 0) "role" ($args.Get 2)) }}
```
Jo3-L, line 30: this won't work properly with fuzzy matching. Say user provides `foo` as their input role and the role that matches is `foobar`. `foo` is not the role name, `foobar` is, but you're using `foo`. Use role ID instead.

util/giveRoleAfterDelay
```diff
@@ -0,0 +1,37 @@
...
+the role: {{ $role.Name }} will be given in {{$args.Get 1}} 
+{{ else }} The user has this role already {{ end }} 
+{{ else }} Role not found, make sure its correct {{ end }}
+{{ else }} {{ $user := .ExecData.user }}
+{{ $role := .ExecData.role }}
+{{ giveRoleName $user $role }}
+role has been given to <@{{$user}}> {{end}}
```
Jo3-L, line 37: dunno if message should be sent after delay, especially if it's long and people forget

Jo3-L (changes requested):
maybe support cancelling giving the role and override past invocations using `scheduleUniqueCC` rather than `execCC`.

JordiHakimZiyech:
FRICK FILE NAME, SHIT

PR #140, opened 2020-12-31
Allow year to be optional in the birthday command

### LINKED ISSUE #139: request: Allow year in birthday command to be optional somehow
This originally came up in the help channels with a user asking if they could not specify their birth year just because they weren't comfortable doing so. 

I considered the fact that since the underage user filter is optional and you don't need the year to check if its the users birthday that perhaps a setting to make the year optional would help.
 
Since the command uses newDate to turn the date provided into time.Time. Maybe a default year (just a placeholder) could work?

ShadowNight573:
Added $yearOptional as a true/false variable and slightly changed how $year is parsed. Default year is 2000 to allow February 29 as a valid date.

Works as far as I tested.
Inverted mode works, setbday works, and mybday works.

- [ x] Code changes have been tested on an instance of YAGPDB, or there are no code changes

- [ x] I have read and followed the [contribution guide](../CONTRIBUTING.md)

Resolves #139

### DIFF
--- a/birthday/README.md
+++ b/birthday/README.md
@@ -6,6 +6,7 @@ All commands can be used with bday or birthday. Example: -getbday or -getbirthda
 This code will also kick/ban users if they are under 13 years old, if you want it to.<p>
 Change ONLY the user variables<p>
 $mods is the list of the Roles IDs that should be able to use all commands.<p>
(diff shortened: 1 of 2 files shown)


Jo3-L:
LGTM, thank you! Happy New Year 🎉

PR #150, opened 2021-01-03
Update afk.go.tmpl

DaviiD1337:


afk/afk.go.tmpl
```diff
@@ -11,16 +11,15 @@
...
 {{/* CONFIGURATION VALUE ENDS HERE */}}
 
-{{/* Dont edit below if you don't know what you're doing ! */}}
 {{ $time := 0 }}{{ $afk := dbGet .User.ID "afk" }}{{ $prefix := index (reFindAllSubmatches `.*?: \x60(.*)\x60\z` (execAdmin "Prefix")) 0 1 }}
 {{ if $args := .Args }}
     {{ if eq (lower (index $args 0)) (print $prefix "afk") }}
         {{ if and (ge (len $args) 2) (not $afk) }}
             {{ $args = (slice .Args 1) }}
-            {{ $message := "" }}
+            {{ $message := "No reason setted" }}
```
Olde7325, line 22: A more sensible afk reason when none is set:
```suggestion
            {{ $message := "afk" }}
```

afk/afk.go.tmpl
```diff
@@ -43,7 +42,9 @@
             {{ with and $duration (toDuration $duration) }} {{ $parsedDur = . }} {{ end }}
             {{ if $parsedDur }}
                 {{ dbSetExpire .User.ID "afk" $message (toInt $parsedDur.Seconds) }}
-            {{ else }} {{ dbSet .User.ID "afk" $message }} {{ end }}
+            {{ else }} 
+                {{ dbSet .User.ID "afk" $message }} {
+            {{ end }}
```
Olde7325, line 47: This doesn't look right
```suggestion
            {{ else }} 
                {{ dbSet .User.ID "afk" $message }}
            {{ end }}
```

afk/afk.go.tmpl
```diff
@@ -11,16 +11,15 @@
 */}}
 
 {{/* CONFIGURATION VALUE STARTS HERE */}}
-{{ $removeAfkOnMessage := true }}{{/* Remove the AFK database from an user if he/she types something in chat. */}}
+{{ $removeAfkOnMessage := false }}{{/* Remove the AFK database from an user if he/she types something in chat. */}}
```
Olde7325, line 14: I would let this be the default setting:
```suggestion
{{ $removeAfkOnMessage := true}}{{/* Remove the AFK database from an user if he/she types something in chat. */}}
```

Olde7325 (changes requested):
Looking good so far, just a few things.

afk/afk.go.tmpl
```diff
@@ -61,7 +62,7 @@
             {{ if gt .ExpiresAt.Unix 0 }} {{ $eta = humanizeDurationSeconds (.ExpiresAt.Sub currentTime) | printf "*%s will be back in around %s.*" $user.Username }} {{ end }}
             {{ sendMessage nil (cembed
                 "author" (sdict "name" (printf "%s is AFK" $user.String) "icon_url" ($user.AvatarURL "256"))
-                "description" (joinStr "\n\n" $eta .Value)
+                "description" (print $eta "\n\nReason: **" .Value "**")
```
jo3-l2, line 65: Honestly not a super big fan of this change, but if others like it better than the original, that's okay.
However, I don't think we should be providing a default message unless strictly necessary. Here's how I would format the description:

* If there is both an ETA and reason, prefer `<ETA>\n<REASON>`
* If there is only an ETA, prefer `<ETA>`
* If there is only a reason, prefer `<REASON>`
* If there is neither an ETA nor a reason, prefer `*No reason provided*`

afk/afk.go.tmpl
jo3-l2: I don't think we should be using a default unless strictly necessary. See review comment later on.

afk/afk.go.tmpl
jo3-l2: Yes, reverting this change would be preferable. From feedback on the first version of AFK, many users were used to the AFK clearing when they sent a message.

jo3-l2 (changes requested):

afk/afk.go.tmpl
jo3-l2: Agreed.

afk/afk.go.tmpl
jo3-l2: I think we shouldn't have a default message at all unless absolutely necessary, see my last review comment.

afk/afk.go.tmpl
Olde7325: Fair enough, couldn't made any sense out of the abbrevation "ETA", thus not commenting on this change.

afk/afk.go.tmpl
jo3-l: Why was this resolved without any changes?

jo3-l:
Thanks!

PR #152, opened 2021-01-12
Fixed the activation trigger description

Skyflash:

**Status**

### DIFF
--- a/afk/afk.go.tmpl
+++ b/afk/afk.go.tmpl
@@ -5,7 +5,7 @@
 	-afk <message>
 	-afk <message> -d <duration>
 
-	Recommended trigger: Regex trigger with trigger `.*`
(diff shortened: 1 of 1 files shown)


Olde7325:
In what way is this beneficial?
You introduce a breaking change, because this custom command is supposed to check if a mentioned user is "afk", and remove the afk status once an afk user sends a message.
With a command trigger all this won't be given.

WickedWizard3588:
Yep, I agree with Olde, just changing the name won't work. Change the code, do all the sorts of need things. Read the code and understand.

Olde7325:
I honestly don't think it'd be a great idea to change the trigger and trigger type, as this would require a rather large refactor including an abstraction of the message listener into a separate CC, using up one CC slot more. 
In my opinion the CC as it is right now is the most efficient way to do it.

PR #158, opened 2021-02-01
cc2file - Medium update

MatiasMFM2001:
## **Please describe the changes this PR makes and why it should be merged:**
This update brings some features:
- Ability to limit execution to 5 CCs
- Ability to show/hide the info of each CC, in message's content and/or attachment's content (it's inserted as a comment)²
- Ability to blacklist CCs to maintain its code private¹

And some minor changes:
- `sendLongMessage` template now uses an `sdict` rather than a `cslice`
- Heavily-commented code: This is mostly for people who has to troubleshoot my code, if there are problems in a future

## **Status**
I didn't really test every single combination of options, but yeah, it works ;)

Some improvements that can be done:
1. Allow some users/roles to override blacklist?
2. Parse `$firstLine`'s content, extract CC info and print it in a format similar to this repo's Leading comments. Something like:
```go
//Before:
{{/* CC INFO: #20 - Contains: `viva` - Case sensitive trigger: `false` - Group: `Respuestas` */}}

//After:
{{/*
	Description:
	Usage:³

	Trigger type: Contains
	Recommended trigger: "viva"
	Case sensitive trigger: "false"

	Credits:
*/}}
```
- In that case, I'll need help with Regex. I'll probably ask in YAG's support server
3. In exported code, if `parseArgs` is used at least one time, copy the `Custom usage text when invalid args provided` argument to the `Usage:` field?
- An easier alternative is replace ` - ` with `\n\t`, so basic output now has some indentation. Something like:
```go
//Before:
{{/* CC INFO: #20 - Contains: `viva` - Case sensitive trigger: `false` - Group: `Respuestas` */}}

//After:
{{/* CC INFO:
	ID #20 
	Contains: `viva`
	Case sensitive trigger: `false`
	Group: `Respuestas`
*/}}
```

### DIFF
--- a/util/cc2file.go.tmpl
+++ b/util/cc2file.go.tmpl
@@ -85,7 +85,7 @@
 
 {{/*Convert each $blacklisted_CCIDs to a RegEx pattern*/}}
 {{range $index, $CCID:= $blacklisted_CCIDs}}
-	{{- $blacklisted_CCIDs.Set $index (printf "^#%d.*" $CCID)}}
+	{{- $blacklisted_CCIDs.Set $index (printf "^#%d -.*" $CCID)}}
 {{end}}
 
 {{/*Process each selected argument in this execution*/}}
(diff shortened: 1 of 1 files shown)


jo3-l:
Thanks for the PR!

MatiasMFM2001:
You're welcome!

I thought of adding your minifier here, but maybe It's not necessary: If you saved your CC (in YAG's CP) is because you're already below the 10K limit. What do you think?

PR #165, opened 2021-02-21
Improved counting

SpecialEliteSNP:
Here's my version of a counting channel. It has built-in support for a -funinfo command which I will PR later. It keeps track of how many times users have counted.

Difference with the already-existing one is that mine doesn't use a role to prevent counting twice in a row. It's all done through the database and all efficient since it's using a single database entry.

It's very customisable using (a lot of) variables. It has a built-in resetcount command which resets the count with a warning as well, so you can't do it on accident. Apart from that the main part is the amount of feedback it gives.


l-zeuch:
I do not see a benefit to the UX when not using a role - the reason most people count in the support server is exactly because of the (admittedly, hoisted) role. Maybe add an option to toggle this? 
Further, we already have a very similar system - Where is the need to add more? How would an end user know which one to use? My experience is that most users who "just want a counting system" are already overwhelmed with the currently existing one...

SpecialEliteSNP:
I understand your point about why people count in the support server, but I've had this same cc for a few months in one of my servers with over 2k members (which are mostly kids). They are still as enthusiastic about it as when I first implemented it. That is without a special role or any other benefits.
I don't think we should question people's motives for wanting to count.

The biggest advantage with this would only really work if I already PR'd my -funinfo command which uses data from this.

SpecialEliteSNP:
If there really are too many variables (which I would totally understand), I could lower them by having come responses just coded in instead.

fun/counting.go.tmpl
```diff
@@ -0,0 +1,102 @@
...
+  {{ else }}
+    {{ print .User.Mention ", you don't have the permissions to reset the count!" }}
+    {{ deleteResponse 5 }}
+  {{ end }}
+{{ else if eq .User.ID $c.l.u }}
+  {{ addReactions "👎" }}
+  {{ deleteTrigger 1 }}
+  {{ print .User.Mention ", " $twice }}
+  {{ deleteResponse 5 }}
+{{ else if or (not (reFind `\A\d+\z` .Message.Content)) (ne $c.l.c (toInt .Message.Content)) }}
```
jo3-l, line 77: Does this allow input like `01`?

fun/counting.go.tmpl
```diff
@@ -0,0 +1,102 @@
...
+      {{ dbSet 20 "counting" $c }}
+      {{ print .User.Mention ", reset the count in <#" .Channel.ID ">!" }}
+      {{ deleteResponse 10 }}
+    {{ end }}
+  {{ else }}
+    {{ print .User.Mention ", you don't have the permissions to reset the count!" }}
+    {{ deleteResponse 5 }}
+  {{ end }}
+{{ else if eq .User.ID $c.l.u }}
+  {{ addReactions "👎" }}
```
jo3-l, line 73: Not a big fan of the reactions, especially given the messages are going to be deleted right after. Seems like a bit of waste and might hit ratelimits if people spam.

fun/counting.go.tmpl
```diff
@@ -0,0 +1,102 @@
...
+  {{ deleteTrigger 1 }}
+  {{ print .User.Mention ", " $twice }}
+  {{ deleteResponse 5 }}
+{{ else if or (not (reFind `\A\d+\z` .Message.Content)) (ne $c.l.c (toInt .Message.Content)) }}
+  {{ addReactions "❌" }}
```
jo3-l, line 78: See previous comment about reactions.

jo3-l (changes requested):

fun/counting.go.tmpl
SpecialEliteSNP: I just checked and yes it does.

fun/counting.go.tmpl
SpecialEliteSNP: I will remove those 👍

fun/counting.go.tmpl
jo3-l: Apologies, wasn't very clear there - I don't believe leading zeros should be allowed. Input like `000001` is typically only ever sent by people who are trolling and attempting to break the system.

SpecialEliteSNP: Added in a check for that

fun/counting.go.tmpl
```diff
@@ -0,0 +1,110 @@
...
+    {{ end }}
+  {{ else }}
+    {{ print .User.Mention ", you don't have the permissions to reset the count!" }}
+    {{ deleteResponse 5 }}
+  {{ end }}
+{{ else if eq .User.ID $c.l.u }}
+  {{ deleteTrigger 1 }}
+  {{ print .User.Mention ", " $twice }}
+  {{ deleteResponse 5 }}
+{{ else if or (not (reFind `\A\d+\z` .Message.Content)) (ne $c.l.c (toInt .Message.Content)) }}
```
jo3-l, line 76: Dunno if this is a good idea but you could check for leading zeros here instead of lower down in the chain - regex `[1-9]\d*` would work fine for preventing leading zeros.

jo3-l (changes requested):
Some last comments before merge -
1) Not a blocker, but I wonder whether having the response delete delay be 10 is a good idea -- that may lead to many bot messages in the chat if someone spams invalid numbers. Not sure though.
2) Might want to recommend that users add a channel override in their command settings for the counting channel so no commands can be ran.

fun/counting.go.tmpl
SpecialEliteSNP: The error message wouldn't really make sense if I do it there. Having a specific error message to why their count was not accepted is better in my opinion. So I'm going to leave it as it is.

SpecialEliteSNP:
I will add that note, but I'm not going to write a whole paragraph on how they should do that.

jo3-l (approved):

jo3-l:
Thanks!

PR #168, opened 2021-03-10
Message to Embed Converter

botagas:
### Short explanation
This is a little command that I made for use in a Discord server. It uses some code from the example commands, but what this does is it identifies the contents of the message, and based on that, it either contains the attachment (image) from the message and the text, or it contains the text only if there are no attachments.
I believe it could be useful if tidied up, to be part of the example custom commands.
### Trigger
The trigger for this custom command is _Regex_ **.***
### Optimization
This might require a lot of changes or there are better ways to do this, but I couldn't find an example that would convert any message to an embed regardless if it was an image, or text, or text and image.
### Prospects
I believe the future of this could be a command that can be used in a private channel to configure in which channels is it used, what emojis are used, and perhaps even scheduling, if we were to go all-in. 

**Status**

### DIFF
--- a/messagetoembed/converter
+++ b/messagetoembed/converter
@@ -1,0 +1,24 @@
+{{ $channel :=  }} {{/* Replace this with your suggestion channel ID */}}
+
+{{if .Message.Attachments}}
+		{{ $embed := cembed
+			"title"  "Photo"
+			"description" (index .Message.Content)
+			"timestamp"  currentTime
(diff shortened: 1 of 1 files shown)


jo3-l:
Thank you for the PR, but personally I feel this is a bit too specific to include here. If you believe otherwise, I'm open to further discussion. Closing for now.

PR #170, opened 2021-03-14
Create Wheel of Fortune.go.tmpl

Hyakki999:

**Status**

### DIFF
--- a/fun/Wheel of Fortune.go.tmpl
+++ b/fun/Wheel of Fortune.go.tmpl
@@ -1,14 +1,11 @@
 {{/*
 	Name: Wheel of Fortune
-	Author: Dr. Hyakkimaru#6020
+	Author: Hyakki999 <https://github.com/Hyakki999>
+	License: MIT
 	
 	Recommended:
 	- Trigger Type: Regex
 	- Trigger: `\A\-(?i)wheel(offortune)?`
-
-	Copyright: 2021 Hyakki999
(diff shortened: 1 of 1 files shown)


fun/Wheel of Fortune.go.tmpl
```diff
@@ -0,0 +1,54 @@
...
+	Name: Wheel of Fortune
+	Author: Dr. Hyakkimaru#6020
+	
+	Recommended:
+	- Trigger Type: Regex
+	- Trigger: `\A\-(?i)wheel(offortune)?`
+
+	Copyright: 2021 Hyakki999
+	License: MIT
+	Repository: https://github.com/Hyakki999/YAGPDB-cc
```
jo3-l, line 11: We have a different copyright notice, so the first part isn't really accurate. I'd also prefer that the repository link points to here (but you may also add something like "also available here on the author's own repo [link]").

As such, I would prefer something like this:
```suggestion
	Author: Hyakki999 <https://github.com/Hyakki999>
	License: MIT
```

jo3-l (changes requested):

fun/Wheel of Fortune.go.tmpl
Hyakki999: Fixed the first part.

jo3-l (approved):

jo3-l:
Thanks!

PR #180, opened 2021-04-11
Small Util Update

devnote-dev:
Additions:
- json converter CC (jsonify)
- ghostping CC v1
- ghostping CC v2

**Status**

### DIFF
--- a/util/ghostping-v2.go.tmpl
+++ b/util/ghostping-v2.go.tmpl
@@ -1,0 +1,47 @@
+{{/*
+    Ghost-Ping Detector CC v2
+
+    Recommended Trigger Type: Regex
+    Recommended Trigger: .*
+    Optional Trigger: <@!?\d{17,19}>
+    
+    Credits:
+    Devonte <https://github.com/devnote-dev>
+*/}}
+
+{{/* THINGS TO CHANGE */}}
+
+{{$CHECK := false}} {{/* Change to "true" for double exec-check */}}
+
+{{$LOG := false}} {{/* Change to "true" to create message logs */}}
+
+{{/* ACTUAL CODE - DO NOT TOUCH */}}
+
+{{if .ExecData}}
+    {{$mentions := ""}}{{$ping := false}}
+    {{if ($m := getMessage nil .ExecData.message)}}
+    {{if $m.Mentions}}{{else}}{{$ping = true}}{{end}}
+    {{else}}{{$ping = true}}{{end}}
+    {{if $ping}}
+    {{range .ExecData.mentions}}{{$mentions = print $mentions "<@" .ID ">, "}}{{end}}
+    {{$col := 16777215}}{{$p := 0}}{{$r := .Member.Roles}}
+    {{range .Guild.Roles}}{{if and (in $r .ID) (.Color) (lt $p .Position)}}{{$p = .Position}}{{$col = .Color}}{{end}}{{end}}
+
+    {{$log := ""}}{{if $LOG}}{{$log = print "\n**Logs:** [Message Logs](" (exec "logs") ")"}}{{end}}
+    {{/* Embed Construct */}}
+    {{$embed := cembed
+        "description" (print "**Channel:** <#" .Channel.ID ">\n**Message:**\n" .ExecData.content "\n\n**Mentioned Users:** " $mentions $log)
+        "color" $col
+        "footer" (sdict "text" "Detector triggered")
+        "timestamp" currentTime}}
+
+    {{/* Content Format */}}
+    {{$msgContent := print "Ghost ping detected from <@" .ExecData.author ">!"}}
+
+    {{sendMessage nil (complexMessage "content" $msgContent "embed" $embed)}}
+    {{else}}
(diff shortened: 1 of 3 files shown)


util/ghostping-v2.go.tmpl
```diff
@@ -0,0 +1,47 @@
...
+
+{{/* ACTUAL CODE - DO NOT TOUCH */}}
+
+{{if .ExecData}}
+    {{$mentions := ""}}{{$ping := false}}
+    {{if ($m := getMessage nil .ExecData.message)}}
+    {{if $m.Mentions}}{{else}}{{$ping = true}}{{end}}
+    {{else}}{{$ping = true}}{{end}}
+    {{if $ping}}
+    {{range .ExecData.mentions}}{{$mentions = print $mentions "<@" .ID ">, "}}{{end}}
```
Pedro-Pessoa, line 26: Same issue as V1: This will always have the message end with a `,`
image

util/ghostping-v2.go.tmpl
```diff
@@ -0,0 +1,47 @@
...
+        "color" $col
+        "footer" (sdict "text" "Detector triggered")
+        "timestamp" currentTime}}
+
+    {{/* Content Format */}}
+    {{$msgContent := print "Ghost ping detected from <@" .ExecData.author ">!"}}
+
+    {{sendMessage nil (complexMessage "content" $msgContent "embed" $embed)}}
+    {{else}}
+    {{if $CHECK}}{{execCC .CCID nil 5 (sdict "message" .Message.ID "author" .Message.Author.ID "mentions" .Message.Mentions "content" .Message.Content)}}{{end}}
```
Pedro-Pessoa, line 43: Same issue as V1: this is an infinite loop.

util/json.go.tmpl
```diff
@@ -0,0 +1,57 @@
...
+
+{{$a := parseArgs 2 "```json <Channel:Mention> <Message:ID>\njson <Channel:Name> <Message:ID>\njson <Channel:ID> <Message:ID>\njson 0 <Message:ID>\n```**Optional Flags:**\n```\n[-j strict-format:flag]\n[-file/f attachment:flag]```"
+(carg "string" "chan") (carg "int" "msgID") (carg "string" "flag: -j | -f")}}
+{{$type := dict 0 "Default Message" 6 "Pinned Message" 7 "Join Message" 8 "Boost Message" 9 "Teir 1 Boost Message" 10 "Teir 2 Boost Message" 11 "Teir 3 Boost Message" 12 "Followed Channel Message" 19 "Reply Message"}}
+{{$struct := "Unknown"}}{{$ver := "JSONify v4.15"}}{{$fa := false}}{{$chan := .Channel}}
+{{$link := joinStr "/" "https://discordapp.com/channels" .Guild.ID .Channel.ID ($a.Get 1)}}{{$ctx := ""}}
+
+{{$l := sendMessageRetID nil (cembed "description" "Converting Message... <a:loading:760219029620523008>")}}
+{{if ne ($a.Get 0) "0"}}
+    {{$chan = getChannel (or (reFind `\d+` ($a.Get 0)) ($a.Get 0))}}
```
Pedro-Pessoa, line 21: This should check if the `$chan` input is valid, otherwise the code will error out.
image

Also, I think `nil` should be accepted as an input, as it is commonly used on CC's.

util/json.go.tmpl
```diff
@@ -0,0 +1,57 @@
...
+    {{$link = joinStr "/" "https://discordapp.com/channels" .Guild.ID $chan.ID ($a.Get 1)}}
+{{end}}
+{{if ($msg := getMessage $chan.ID ($a.Get 1))}}
+    {{with $msg.Embeds}}{{$struct = print (title (index . 0).Type) " Embed"}}{{else}}
+    {{$struct = or (and $msg.Attachments "Attachment Message") (and (eq $msg.Type 0 6 7 8 9 10 11 12 19) ($type.Get (toInt $msg.Type))) $struct}}
```
Pedro-Pessoa, line 26: I don't think this `eq` check is necessary, since `$type.Get` will be false if the type msg is none of those.

Pedro-Pessoa (changes requested):
Besides all the comments I left, I'm not so happy with the formatting of the ghostping CCs.
They have plenty of space left and I think they could be formatted better to improve readability.
On the v1, lines 20 to 22, for example. That `else` looks like it s being checked against `{{if not $m.Mentions}}` when in fact it is being checked against `{{if ($m := getMessage nil .ExecData.message)}}`.

I don't think it is necessary to format the code like this here. New lines are free and so are idents.

Otherwise, everything looks good to me.

util/json.go.tmpl
devnote-dev: The commit states that `0` is the placeholder for the triggering channel, not `nil` literally.

Pedro-Pessoa: I know. That's why I said: `Also, I think nil should be accepted as an input`. This is just a suggestion.
But, the code erroring out like this is not good at all. It will error out if any invalid channel is provided. `nil` was just an example.
image
This error should be handled just like the invalid message error is handled. 
image

devnote-dev: Fixed in the most recent commit.

Pedro-Pessoa (approved):

Pedro-Pessoa:
This looks good now.
I'm not a fan of the formatting though, but it can be changed later.

PR #181, opened 2021-04-13
New RAID GUARD System

Engineer152:
This code was created in response to a raid that was done on a server that I am an admin. Over 100 people raided the server, and started sending messages, DMing people. It took us admins at least 30 min to realized that all these people were from all one raid, as we hadn't been pinged, or notified. That's when I came up with this idea of a monitoring system. This system will oversee all incoming members, if that member has an account that is younger than a certain age (1 day default), then they will be added to a database list. If that list reaches a certain amount of people (25 ppl default) within 10 min, then it will send a notification to the modlog. This notification will alert the admin (or a role of your choosing), and ask if they would like to kick/ban all of the RAID members at once. An option has also been  provided to manually reset the raid list. The program will run through the entire list and remove all raiders for good.

**Status**

### DIFF
--- a/util/Raid Guard/raidadmin.go.templ
+++ b/util/Raid Guard/raidadmin.go.templ
@@ -1,0 +1,106 @@
+{{/*
+	Trigger: raid
+	Trigger Type: Command
+
+	About: This code was created in response to a raid that was done on a server that I am an admin. Over 100 people raided the server, and started sending messages, 
+	DMing people. It took us admins atleast 30 min to realized that all these people were from all one raid, as we hadnt been pinged, or notified. Thats when i came up with
+	this idea of a monitoring system. This system will oversee all incoming members, if that member has an account that is younger than a certain age (1 day default), then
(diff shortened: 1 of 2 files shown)


l-zeuch:
A few things missing:
1. Where is the mentioned role to be configured?
2. Would recommend adding a `README.md` as documentation, due to your PR being a system of commands, as per our contributing guidelines...
3. Could you indicate where the configuration areas are pointed out as well as where the code starts? Something like
```
{{/* Config Area start */}}
[...]
{{/* Actual code - do not touch */}}
```

This would make it a lot easier for potential users to install and set up.

l-zeuch (changes requested):
A few things missing.. see my first comment

Engineer152:
I have now updated the formatting for the Join command trigger to be more user friendly. As well as I have added a Readme.md

l-zeuch (approved):
Thanks!

PR #182, opened 2021-04-13
Update and move RAID GUARD files from util to Moderation.

Engineer152:
This was just a quick update to the join file, and moved all files to the moderation folder.


l-zeuch (approved):
I think we've covered everything on the server...
Thanks again!

PR #195, opened 2021-04-26
Update starboardListener.go.tmpl

dvoraknt:
Added auto pruning of DB value when character count exceeds 90k to prevent hitting 100kb DB storage limit. Since this starboard stores user ID's of every member who reacts it's likely that large servers will eventually reach the DB size cap. Added to the starboard listener as it would have put starboard.go.tmpl over CC character count. Naturally I was unable to test this at scale but it performed flawlessly on removals up to 5k characters in testing. All without adding a single DB interaction :ye:

**Status**


l-zeuch (approved):
LGTM!

PR #207, opened 2021-05-29
chore(bigemojiv2,staffOnDuty): update regex

EsenDR:
Simple change, just updating regex for both.

**Status**

### DIFF
--- a/moderation/staffOnDuty.go.tmpl
+++ b/moderation/staffOnDuty.go.tmpl
@@ -1,6 +1,6 @@
 {{/*
 	Trigger type: regex 
-	Trigger: `\A-o(n|ff)duty` (triggers on `-onduty` and `-offduty`)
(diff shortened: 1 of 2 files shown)


l-zeuch (approved):
Esen tryna getting top-contributor

dvoraknt:
what? this breaks both of them as one requires arguments and the other has optional arguments. adding (\s+|\z)(\s+|\z) makes bigemoji completely nonfunctional and breaks half of the on duty system. The yag ping is an understandable alternative addition but whitespace/endline need to be removed

l-zeuch:
It does not. The trigger is directly taken from the `;rgalias` snippet, which we'd know if it does not work as you described.

jo3-l:
> If nothing else, (\s+|\z)(\s+|\z) is incorrect, as it requires at least 2
spaces or EOF.

dvoraknt:
I didn't explain well enough before the edit. Requiring two spaces between trigger and argument is silly

l-zeuch:
This at max - apparently it slipped? dunno. The snippet at least is correct

PR #208, opened 2021-06-02
Create transferroles.go.tmpl

2x2master1:
This is my 2nd PR here. I worked on a command that makes YAGPDB give one member all roles another member has. (I only used the sleep function in line 38 because it was really buggy without it.) If I should not respond to a comment, please ping me on discord! Thanks for your time.

It adds a new custom command, as seen above.

**Status**

### DIFF
--- a/util/transferroles.go.tmpl
+++ b/util/transferroles.go.tmpl
@@ -1,0 +1,55 @@
+{{/*
+	Custom Command to transfer a member's roles to another member
+
+
+	Trigger type: Regex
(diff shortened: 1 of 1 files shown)


jo3-l:
Thank you for the PR, but I don't want to accept this in its current state due to the fact that it makes requests one-by-one (meaning it would require X requests to change X roles).
Instead, I'd like to wait for https://github.com/jonas747/yagpdb/pull/750 to land on the main bot code and then we can rewrite this to make 1 single request overwriting the roles.

2x2master1:
Makes sense. Thanks!

PR #224, opened 2021-07-06
refactor(*): reorganize files 🧹

### LINKED ISSUE #38: request: allow usage of attached images as base for meme CC
Allow the usage of Attached images as the base for the meme instead of their link. If a attachment is found the command will require only top text as arg.
Also make the Bottom Text optional, and if is not provided use **Bottom Text** as default ~~because this is a meme by itself~~

### LINKED ISSUE #194: Formatting trigger in code
i have seen several users copy the trigger with ` so maybe removing that or using a more elegant way mighz be better.
In this example i used the -be cc

### LINKED ISSUE #222: proposal: remove instructions in leading comments completely
Now that we have the website and associated infra set up (we have a job making sure the website builds and a job automatically deploying the website on new commits to `master`), I propose that instructions be removed from leading comments completely. All that would be left would be 1) a link to the associated page on the website and 2) information about the author of the CC.

**Motivation**
I think it goes without saying that the website is a far better place to browse and view custom commands than this GitHub repository. The question arises as to why exactly we need to keep instructions in leading comments anymore. Personally, I don't see any benefit of doing so, but I do see several downsides. In particular, if we want to keep the website up to date with the code here, we currently need to change up to 4 things in a single PR:
1) We need to change the actual code.
2) We need to change the instructions in the leading comment.
3) We need to change the code on the website. 
4) We need to change the instructions on the website.

Only 2 of those 4 things are really necessary, the rest is just repetitive work. If we remove the need to keep instructions in header comment, that's only 3 things to change at max, rather than 4. To me, that outweighs any benefit keeping instructions in leading comments could bring.

What do we think of this?

jo3-l:
**Summary**

* Shorten leading comments. No more detailed instructions, just a short description + link to website page + author.
* Move custom command code to `src/`.
* Change custom command filenames to `snake_case`, to be consistent with Go.
* Don't duplicate code on the website. Instead, include the content from the file using a plugin.
* Rewrite pages on the website to make use of Docusaurus features like line highlighting, admonitions, and so on. Also use a consistent style across all pages.
* Probably some more things I've forgotten

**Progress**

* [x] Move custom command to `src/`
* [x] Change custom command filenames to `snake_case`
* [ ] Replace <LINK> placeholders in leading comments with actual links
* [ ] Rewrite website pages
  * [x] `afk/`
  * [x] `birthday/`
  * [x] `code-snippets/`
  * [x] `fun/`
  * [x] `birthday/`
  * [x] `info/`
  * [x] `leveling/`
  * [ ] `moderation/`
  * [ ] `suggestion/`
  * [ ] `utilities/`
  * [ ] `tags/`
* [x] Shorten leading comments
  * [x] `afk/`
  * [x] `birthday/`
  * [x] `code-snippets/`
  * [x] `fun/`
  * [x] `birthday/`
  * [x] `info/`
  * [x] `leveling/`
  * [x] `moderation/`
  * [x] `suggestion/`
  * [x] `utilities/`
  * [x] `tags/`

Closes #222 
Closes #194
Closes #38 (fix for this was included in this and I'm too lazy to move it out)


l-zeuch:
Joe you mentioned in #222 to also switch the branching model, aka make those changes also appear in contributing guidelines

I'll let you know what else you probably forgot

jo3-l:
switching the branching model isn't a concern of this PR

l-zeuch:
aight :ye:

PR #238, opened 2021-09-23
New code-snippet remove_repeated_characters

LattandiFacundo:
Adding new command.


2x2master1:
I do believe that this CC would fit better into the snippets category with a few adjustments.

LattandiFacundo:
Yeah, better. I'll change the category.
What adjustments do you think this needs?

2x2master1:
I'd say make the code not based on a message that is sent, but on a variable instead. Line 8 can be a user-defined variable, and the last line can be removed entirely as there is no need to send a message.

jo3-l (changes requested):
Mostly small issues, overall the command looks fine.
Please change the filename to `remove_repeated_letters.go.tmpl` - we use snake_case for code filenames.

Perhaps it would be useful to show an example of having different thresholds for different letters, so the `hello => helo` issue could be avoided by setting the threshold for `l` to `2`. Here is a quick example:
```go
{{ $ctr := -1 }} {{ $c := "" }}
{{ $cleaned := "" }}
{{ range $e := $splitted }}
  {{- if eq $e $c }}
    {{- if gt $ctr 0 }} {{- $cleaned = print $cleaned $e }} {{- end -}}
    {{- $ctr = sub $ctr 1 }}
  {{- else }}
    {{- $c = $e }}
    {{- if eq $c "l" }} {{- $ctr = 1 }} {{- else }} {{- $ctr = 0 }} {{- end -}}
{{ end }}
```
`$ctr` tracks how many characters of the current type are allowed before further occurrences are removed.

website/docs/code-snippets/removeRepeatedLetters.md
```diff
@@ -0,0 +1,29 @@
...
+English words like `Hello` will return `Helo`
+Take this in mind if you want to use the code in another commands, when listing words to match.
+
+## Confguration
+
+- Change the sentence to arrange in `$toFix`
+
+## Code
+
+```gotmpl file=../../../src/utilities/removeRepeatedLetters.go.tmpl
```
jo3-l, line 23:
```suggestion
```gotmpl file=../../../src/code_snippets/removeRepeatedLetters.go.tmpl
```

LattandiFacundo:
Your idea is great but this command is not for this, I mean, it is easier to list Hello as Helo that list all words with double and single letters ("hheelloo"), **this command is to make a template able to detect some words**, not to answer nothing.

LemmeCry:
Snippets are pieces of code that showcase what can be done and how it can be achieved. With that in mind, showing more possibilities and variations using threshold only makes sense for a user because their use-case might be different. If this were a command, no such additional functionality would be needed, but being a snippet, it's not meant to function as command, but rather, serve as a basis for a user's project where they can determine how the concept covered in the snippet will be incorporated in it.

LattandiFacundo:
### To explain me better

**With out your fix**
For input: `heeellllooooo`
Expected output: `helo`
For input: `hello`
Expected output: `helo`

Here you can easily detect all `helo` in a sentence.

**With your fix**
For input: `heeellllooooo`
Expected output: `heelloo`
For input: `hello`
Expected output: `hello`

Here you can not detect `hello`, you would need to write all posible variations `("hello", "helloo", "heelloo", etc.)`.

l-zeuch:
It doesn't matter what you'd need to write or not to write.

What matters is that snippets are more an example on how to achieve something, meaning it is desirable to include more cases, such as the requested one of showcasing a threshold required to reach before characters are removed.
Only showcasing one specific case makes me not want to merge this MR, as I believe this snippet would then lack covering a not uncommon issue.

As LemmeCry already pointed out, it is not a fully featured custom command you're requesting into this repository, but a snippet -- snippets are examples, which definitely should showcase covering edge-cases.
There is nothing wrong with writing another block of code and documenting both appropriately. Users browsing snippets will know how to read them.
But straight up not including an edge-case due to reasons I personally don't quite understand is in my opinion a blocker.

Kind Regards

LattandiFacundo:
Fine! I will re-write the documentation in some hours to make it an snippet.

I think I do well the path changing?

LattandiFacundo:
I decided to publish this solution because I was like a complete morning wondering how to do this.
And I published it as a command because I did not knew about splitters, but as I say, I will re-write the documentation.

jo3-l:
Hi @LattandiFacundo, I think you've misunderstood a bit. Consider your example of `heeellllooooo`. I think it should become `hello`, *not* `heelloo` - the threshold for `l` is 2 but it is 1 for all other characters. The reason this would be useful would be to differentiate between `hello` and `helo`. (In this specific case it doesn't matter, but say we wanted to find `oops` in the string. Without an override for `o` it would be collapsed to `ops` which is a completely different word.)

LattandiFacundo:
If you think this is a good addition I am going to think a good form to add this to my existing idea.

jo3-l (changes requested):
The latest commit modifying the implementation seems to have introduced several regressions.
- `o` is transformed to `oo`.
- `oz` is transformed to `oo`.

src/code_snippets/remove_repeated_characters.go.tmpl
```diff
@@ -0,0 +1,28 @@
+{{/*
+    Removes the repeated letters in a text like 'byyeeee' => 'bye'
+    Useful to add it in another more complex commands.
+
+    Author: Faculord <https://github.com/LattandiFacundo>
+*/}}
+
+{{ $toFix := "oooooooooooopppppppppppssssssss"}}
+{{ $exception := "o"}} {{/*Would keep two repeated characters if they matches the exception*/}}
```
jo3-l, line 9: I don't mean to be pedantic, but - for a snippet - I don't feel that this is adequate. Snippets should be fairly easy to customize. For example, what if I wanted to have two exceptions? What if I wanted one character to have a limit of 3 and the other 2? (You may disagree, of course; let me know if you don't think what I'm saying makes sense.)

website/docs/code-snippets/remove_repeated_characters.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Remove repeated characters
+---
+
+Removes repeated characters in strings.
```
jo3-l, line 5: Perhaps consider including an example.

website/docs/code-snippets/remove_repeated_characters.md
```diff
@@ -0,0 +1,19 @@
...
+Useful addition to commands that needs to compare words.
+
+## Usage
+Replace `$toFix` and `$exception`.
```
jo3-l, line 9: A bit more explanation would be nice here.

jo3-l (changes requested):

LattandiFacundo:
Hello, I will add a $exeptions with a list of character exeptions, but I do not know how to make it so for each exeption you have a threshold. Also, I am so busy with school now. I do not forget about this but I will fix this later.

LattandiFacundo:
After posting this I found a way to make this. I will post the fix in some hours.
About documentation, I am a little bit confused, some days ago someone comment that I have to make it shorter and with out examples, I have no problems to add them, but, should I?

jo3-l:
@LattandiFacundo no worries, I have school too so I totally understand if you're busy 😄 Please don't feel rushed to finish this PR in any way, feel free to take your time if you have other things to do :)

Custom thresholds: I think I posted a working example above. Let me know if you have trouble implementing it and I can help.

Examples: I would personally like to see more examples (refer to the other snippets' pages).

LattandiFacundo:
If there is something to correct let me know

src/code_snippets/remove_repeated_characters.go.tmpl
```diff
@@ -0,0 +1,46 @@
...
+			{{- $remaining = (toInt ($exceptions.Get $e)) }}
+			{{- $fixed = print $fixed $e }}
+			{{- $catchedException = true }}
+		{{- else if and (eq $e $exception) (ge $remaining 1) }}
+			{{- $fixed = print $fixed $e }}
+			{{- $catchedException = true }}
+		{{- else if ne $before $e}}
+			{{- $catchedException = false }}
+		{{- end }}
+	{{- end }}
```
jo3-l, line 32: Can you explain what this does, and why `.Get` isn't sufficient?

LattandiFacundo: The firts one is for the first catch of the exception, in the second one when it print but do not actualize the remaining and the other one to let the default threshold work when there is no exception after exeptions range.

jo3-l: That still doesn't quite answer why `.Get` isn't sufficient. The reason I'm concerned about this is that `.Get` is faster than a loop through the map (which could cause a slowdown with large input + exceptions). Also, it's definitely possible using `.Get` (see https://github.com/yagpdb-cc/yagpdb-cc/pull/238#issuecomment-925977801) and likely also cleaner.

jo3-l:
This has been stale for a while, so closing. If you find time to work on this again feel free to re-open the PR. Thanks!

PR #240, opened 2021-11-19
Ticket channel cc

mrhappyma:
**Status**

- [X] I have [written documentation](../WRITING-DOCUMENTATION.md) for my changes, or there is no need to


src/utilities/ticket-channel.go.tmpl
```diff
@@ -0,0 +1,13 @@
+{{/*
+	Takes all messages sent in a specified channel, creates a ticket owned by the user who sent it and witht the message as the reason, and deletes the original message.
```
l-zeuch, line 2:
```suggestion
	Creates a ticket owned by the user who sent it and with the message as the reason, and deletes the original message.
```
Please adhere to the correct spelling. Moreover, this code does **not** take all messages in a channel, so please remove that.

Yes, the trigger causes this behaviour, but it's not the code. The leading comment shall describe the code in a neutral manner, with no regards to the trigger, or trigger type.

Moreover, this phrasing could cause confusion for end-users, as it might imply that it **moves** (as in, take elsewhere) all messages in that channel. Consider rephrasing it to be more precise.

website/docs/utilities/ticket-channel.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Ticket Channel
+---
+
+Takes all messages sent in a specified channel, creates a ticket owned by the user who sent it and witht the message as the reason, and deletes the original message.
```
l-zeuch, line 5:
```suggestion
Creates a ticket owned by the user who sent it and with the message as the reason, and deletes the original message.
```
See my above comment.

website/docs/utilities/ticket-channel.md
```diff
@@ -0,0 +1,32 @@
...
+
+Make sure you have the tickets system set up first.
+
+:::
+
+## Trigger
+
+**Type:** `Regex`<br />
+**Trigger:** `\A`<br />
+**Additional options:** Be sure to restrict it to only run in a specified channel
```
l-zeuch, line 17: Please put more stress on this, as it is more a requirement than an option.
Consider using a warning block.

l-zeuch (changes requested):
In what way is this useful to a potential user?
This is quite trivial code and does nothing more than the already inbuilt functionality of `ticket create`. As such, I'm probably going to close this pull request.
If you still wish that it's being merged, consider adding more features to it.

In either case, I request the following changes.

l-zeuch:
Please also adhere to our [Naming Conventions](https://github.com/yagpdb-cc/yagpdb-cc/blob/master/CONTRIBUTING.md#naming-conventions), as per CONTRIBUTING:
> - Use `snake_case` for folders and for files: `foo_bar/baz_buz.go.tmpl`, not `fooBar/bazBuz.go.tmpl`.
> - Use the file extension `.go.tmpl`, for _Go template_.

mrhappyma:
For me this is useful as an alternate way of having users create tickets, where I don't want them to use commands. Though I completely understand if you close this since it is a bit of a special use-case cc.

I'll get right on adding those suggestions (or you can just close this before I get to that lol).

website/docs/utilities/ticket-channel.md
mrhappyma: Warning block instead of, or in addition to?

l-zeuch:
Closing.

As this is simply replicating base functionality without further added features, it is not needed.

Thank you for your time.

PR #241, opened 2021-11-21
Update afk.go.tmpl

BlackWolfWoof:

When using " as prefix it would cause an issue. This patch should fix that. 
It is not really a good patch as it doesn't solve the root problem of it but it still works. Thats why i am not pushing it and just creating a pr for someone else to review.

**Status**

- [x] I have [written documentation](../WRITING-DOCUMENTATION.md) for my changes, or there is no need to

### DIFF
--- a/src/afk/afk.go.tmpl
+++ b/src/afk/afk.go.tmpl
@@ -11,7 +11,8 @@
 {{/* End of configurable values */}}
 
 {{ $time := 0 }}{{ $afk := dbGet .User.ID "afk" }}{{ $prefix := index (reFindAllSubmatches `.*?: \x60(.*)\x60\z` (execAdmin "Prefix")) 0 1 }}{{ $desc := "" }}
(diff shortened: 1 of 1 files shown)


LemmeCry:
I don't see any changes...?

BlackWolfWoof:
Hmmm great, ctrl v doesn't work on web...

l-zeuch (commented):
As far as I know, this will also break with these prefixes
* ' (Single quotes)
* ` (Backticks)
Seeing as those are also used to denote argument strings.
Unfortunately I'm currently unable to test this hypothesis, but if that is the case, we should consider implementing a similar fallback for those.

jo3-l:
I don't think this is an edge case worth complicating the code for. (Of course, I'm happy to hear arguments to the contrary; did someone using this command come across this problem?)

If we are to fix this issue -- and again, I am unconvinced at the moment that doing so is needed -- I have a different way in mind that is more robust.

BlackWolfWoof:
Yep, someone used it and came to us and said it won't respond.
Also I won't mind if it doesn't get accepted (why did i put 2h into that code aaa) because it is a bit meh the fix xd

l-zeuch:
Closing due to the above.

PR #310, opened 2023-04-12
docs: remove official remark

l-zeuch:
Remove the "official" remark and replace it with the similar, but less
assertive term "vetted".

Add to the "Need help" section that not every staff may know about every
custom command present here.

The rationale behind this change is the fact that "official" implies
that all staff are more or less knowledgeable about the codebase present
in this collection, which is simply not true.
Further, it cannot be reasonably expected that all staff *are* indeed
knowledgeable about this collection.

"Vetted" has a similar meaning, though it is less assertive:
Someone with good knowledge has looked at the code and confirmed it to
be -- to the best of their abilities -- free of malicious code.

Third-party content always has a risk with it, which is also noted down
(though to a different extent) in LICENSE.md. This change should make it
clearer that whilst official staff *looked* at it, it doesn't mean that
we also serve a guarantee for it.

Signed-off-by: Luca Zeuch <l-zeuch@email.de>

**Status**

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,20 +12,27 @@
 
 # Introduction
 
-A curated collection of custom commands for the [YAGPDB Discord Bot](https://yagpdb.xyz) to enhance your server. It is:
+A curated collection of custom commands for the [YAGPDB Discord Bot](https://yagpdb.xyz) to enhance your server.
+
+It is:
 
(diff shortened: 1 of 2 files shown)


l-zeuch:
@jo3-l request for comment / thoughts.

jo3-l (approved):
No strong feelings either way.

str8UP4hung:
I don't have a clue what this is

PR #323, opened 2024-03-15
Updating src/info channel.go.tmpl and user.go.tmpl

H1nr1:
Updates the src/info/channel.go.tmpl and src/info/user.go.tmpl files to use present-day functions.
For the user command, roles were restructured to be in order of server role list position.
Other small additions and changes listed specifically in each commit.

**Status**

### DIFF
--- a/src/info/channel.go.tmpl
+++ b/src/info/channel.go.tmpl
@@ -11,21 +11,29 @@
 	{{ $channel = $args.Get 0 }}
 {{ end }}
 
-{{ $types := cslice "Text" "DM" "Voice" "Group DM" "Category" "News" "Store" }}
+{{ $types := sdict 
+	"0" "Text" "1" "DM" "2" "Voice" "3" "Group DM" "4" "Category" "5" "Announcement" 
+	"10" "Thread - Announcement" "11" "Thread - Public" "12" "Thread - Private" 
+	"13" "Stage" "14" "Directory" "15" "Forum" "16" "Media"
+}}
+{{ $type := $types.Get (json $channel.Type) }}
 {{ $nsfw := "No" }}
 {{ $parent := "*None set*" }}
 {{ if $channel.NSFW }} {{ $nsfw = "Yes" }} {{ end }}
-{{ with $channel.ParentID }} {{ $parent = printf "<#%d>" . }} {{ end }}
-{{ $createdAt := div $channel.ID 4194304 | add 1420070400000 | mult 1000000 | toDuration | (newDate 1970 1 1 0 0 0).Add }}
-{{ sendMessage nil (cembed
+{{ with $channel.ParentID }}
+	{{ $parent = printf "<#%d>" . }}
+	{{ if (getChannel .).IsForum }} {{ $type = "Thread - Forum" }} {{ end}}
+{{ end }}
+{{ $createdAt := snowflakeToTime $channel.ID }}
+{{ sendMessage nil ($e:=cembed
 	"title" (printf "❯ Info for #%s" $channel.Name)
 	"fields" (cslice
 		(sdict "name" "❯ ID" "value" (str $channel.ID) "inline" true)
 		(sdict "name" "❯ Topic" "value" (or $channel.Topic "*None set*") "inline" true)
 		(sdict "name" "❯ Parent Channel" "value" $parent "inline" true)
 		(sdict "name" "❯ NSFW" "value" $nsfw "inline" true)
-		(sdict "name" "❯ Position" "value" (str (add $channel.Position 1)) "inline" true)
-		(sdict "name" "❯ Type" "value" (index $types $channel.Type) "inline" true)
+		(sdict "name" "❯ Position" "value" (add $channel.Position 1 | str) "inline" true)
+		(sdict "name" "❯ Type" "value" $type "inline" true)
 	)
 	"color" 14232643
 	"footer" (sdict "text" "Created at")
--- a/src/info/channel.go.tmpl
+++ b/src/info/channel.go.tmpl
@@ -26,6 +26,7 @@
 	15 "Forum"
 	16 "Media"
 }}
+{{/* $channel.Type is converted to json, then int as toInt unexpectedly returns 0 for .Channel.Type currently */}}
 {{ $type := $types.Get (toInt (json $channel.Type)) }}
 {{ $nsfw := "No" }}
 {{ $parent := "*None set*" }}
--- a/src/info/channel.go.tmpl
+++ b/src/info/channel.go.tmpl
@@ -11,28 +11,38 @@
 	{{ $channel = $args.Get 0 }}
 {{ end }}
 
-{{ $types := sdict 
-	"0" "Text" "1" "DM" "2" "Voice" "3" "Group DM" "4" "Category" "5" "Announcement" 
-	"10" "Thread - Announcement" "11" "Thread - Public" "12" "Thread - Private" 
-	"13" "Stage" "14" "Directory" "15" "Forum" "16" "Media"
+{{ $types := dict 
+	0 "Text"
+	1 "DM"
+	2 "Voice"
+	3 "Group DM"
+	4 "Category"
+	5 "Announcement"
+	10 "Thread - Announcement"
+	11 "Thread - Public"
+	12 "Thread - Private"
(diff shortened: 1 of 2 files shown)


src/info/channel.go.tmpl
```diff
@@ -11,21 +11,29 @@
 	{{ $channel = $args.Get 0 }}
 {{ end }}
 
-{{ $types := cslice "Text" "DM" "Voice" "Group DM" "Category" "News" "Store" }}
+{{ $types := sdict 
+	"0" "Text" "1" "DM" "2" "Voice" "3" "Group DM" "4" "Category" "5" "Announcement" 
+	"10" "Thread - Announcement" "11" "Thread - Public" "12" "Thread - Private" 
+	"13" "Stage" "14" "Directory" "15" "Forum" "16" "Media"
```
jo3-l, line 17: Newline for each entry, please.
Semantically I think this makes more sense as a dict with int keys, even though that would require an extra conversion `toInt (json $channel.Type)`.

src/info/channel.go.tmpl
```diff
@@ -11,21 +11,29 @@
...
 {{ $nsfw := "No" }}
 {{ $parent := "*None set*" }}
 {{ if $channel.NSFW }} {{ $nsfw = "Yes" }} {{ end }}
-{{ with $channel.ParentID }} {{ $parent = printf "<#%d>" . }} {{ end }}
-{{ $createdAt := div $channel.ID 4194304 | add 1420070400000 | mult 1000000 | toDuration | (newDate 1970 1 1 0 0 0).Add }}
-{{ sendMessage nil (cembed
+{{ with $channel.ParentID }}
+	{{ $parent = printf "<#%d>" . }}
+	{{ if (getChannel .).IsForum }} {{ $type = "Thread - Forum" }} {{ end}}
+{{ end }}
```
jo3-l, line 26: We already have the channel struct so let's not issue a gratuitous `getChannel` call fetching it again here. Just use a normal `if` and access `$channel.IsForum`.

src/info/channel.go.tmpl
```diff
@@ -11,21 +11,29 @@
...
+{{ $createdAt := snowflakeToTime $channel.ID }}
+{{ sendMessage nil ($e:=cembed
```
jo3-l, line 28: Is `$e` used later?

src/info/channel.go.tmpl
```diff
@@ -11,21 +11,29 @@
...
 	"title" (printf "❯ Info for #%s" $channel.Name)
 	"fields" (cslice
 		(sdict "name" "❯ ID" "value" (str $channel.ID) "inline" true)
 		(sdict "name" "❯ Topic" "value" (or $channel.Topic "*None set*") "inline" true)
 		(sdict "name" "❯ Parent Channel" "value" $parent "inline" true)
 		(sdict "name" "❯ NSFW" "value" $nsfw "inline" true)
-		(sdict "name" "❯ Position" "value" (str (add $channel.Position 1)) "inline" true)
-		(sdict "name" "❯ Type" "value" (index $types $channel.Type) "inline" true)
+		(sdict "name" "❯ Position" "value" (add $channel.Position 1 | str) "inline" true)
```
jo3-l, line 35: Please revert this. There is no benefit when there are only two function calls.

src/info/user.go.tmpl
```diff
@@ -21,31 +21,47 @@
...
 
-{{ $roleMentions := cslice }}
-{{ range $member.Roles }} {{- $roleMentions = $roleMentions.Append (print "<@&" . ">") -}} {{ end }}
+{{ $roles := $member.Roles }}
+{{ $rolesDict := dict }} {{ $rolesPos := cslice }}
+{{ range $roles }}
+	{{- $role := $.Guild.GetRole . }}
+	{{- $rolesDict.Set $role.Position $role.ID }}
+	{{- $rolesPos = $rolesPos.Append $role.Position -}}
+{{ end }}
```
jo3-l, line 30: You can sort a slice of `sdict "role" [role object] "position" [position]`, providing `sort` with the option `key=position` instead of building `$rolesPos` and then recovering the roles via `$rolesDict`.

src/info/user.go.tmpl
```diff
@@ -21,31 +21,47 @@
...
+	{{ $rolesSorted = slice $rolesSorted 0 $ROLE_TRUNCATION_THRESHOLD }}
 {{ end }}
+{{ $rolesPos = $rolesSorted }}
 
-{{ $roleDisplay := joinStr ", " $roleMentions.StringSlice }}
+{{ $roleMentions := cslice }}
+{{ range $rolesPos }}
+	{{- $roleMentions = $roleMentions.Append (printf "<@&%d>" ($rolesDict.Get .)) -}}
+{{ end }}
+{{ $roleMentions = joinStr ", " $roleMentions.StringSlice }}
```
jo3-l, line 44: You can leave out the `StringSlice`.

src/info/user.go.tmpl
```diff
@@ -21,31 +21,47 @@
...
 	"fields" (cslice
-		(sdict "name" "❯ Nickname" "value" (or $member.Nick "*None set*"))
-		(sdict "name" "❯ Joined At" "value" ($member.JoinedAt.Parse.Format "Jan 02, 2006 3:04 AM"))
-		(sdict "name" "❯ Created At" "value" ($createdAt.Format "Monday, January 2, 2006 at 3:04 AM"))
-		(sdict "name" (printf "❯ Roles (%d Total)" (len $member.Roles)) "value" (or $roleDisplay "n/a"))
+		(sdict "name" "❯ Created At" "value" (printf "<t:%d:F>" $createdAt.Unix))
+		(sdict "name" "❯ Joined At" "value" (printf "<t:%d:F>" $member.JoinedAt.Parse.Unix))
+		(sdict "name" "❯ Names" "value" (printf "Global: %s\nNick: %s" 
+			(or $user.Globalname "*None set*") (or $member.Nick "*None set*")
+		) )
```
jo3-l, line 63: No space between parens, for consistency with the rest of the file.

jo3-l (changes requested):
Some nits since this is a stylistic change.

src/info/channel.go.tmpl
H1nr1: Should I add a comment as to why we convert to json, then int?
It seems like a waste, but is needed to get a comparable value type that preserves the value with how `.Channel.Type` is currently returned.

src/info/channel.go.tmpl
H1nr1: I did not think of the case where the channel given is the forum channel itself. I was thinking one of the forum thread channel would be the channel to retrieve info on.
In that case, `getChannel` is needed if the channel is a forum thread, as `.IsForum` returns false for forum threads (true for `.IsThread`); but true for the forum channel (.ParentID) itself.
I can add conditionals to determine whether the channel is a forum or forum thread as `.IsForum` returns differently for the forum channel itself and a forum thread. I think both should present the channel type as "Thread - Forum"; or maybe we separate them into "Forum" and "Forum - Thread". What do you think?

src/info/channel.go.tmpl
H1nr1: This was left-over from testing, sorry about that; will remove.

src/info/channel.go.tmpl
H1nr1: Was done for readability, but has been reverted.

src/info/channel.go.tmpl
jo3-l: Thanks, good catch. I read the code too hastily and misunderstood your intent. It is fine as is.

src/info/channel.go.tmpl
jo3-l: Ideally `toInt` is fixed so that `toInt .Channel.Type` gives the expected result. When that happens, we can change this to just `toInt .Channel.Type`. But for the time being, a comment would be useful.

src/info/user.go.tmpl
H1nr1: I see, very good then. I read the docs on `sort` to see the best way about it but did not understand fully what it meant. Will change soon!

jo3-l (approved):

jo3-l:
Thanks.

PR #334, opened 2024-07-20
counting: add counting v2 command

H1nr1:
This version of a counting command allows for the usage of base 10 and roman numerals, optionally joined by the usage of math functions and/or operators to count from 1, incrementing by 1, until failure.

With a range of configurable variables, the user can chose to have the ability to count twice, or more, in a row, roles for correct counting users and incorrect counting users, effectively allowing a blacklist; saves, reactions, and more!

Also bundled in this command is a statistics command, with optional arguments to view server or user stats individually or a leaderboard for users which have counted correctly the most times.

Possible changes/request for help deciding
- As the character count is nearly 10k, even the configuring of variables may push it over. Where would be best to slim down without losing readability? I include quite a bit of space around and within comments which could be removed. The comments to help the user configure the variables could possibly be removed, or at least shortened, given they are described in the command docs. I would rather not ask the user to remove anything themselves; for example, the header comment, as I've seen elsewhere. Splitting the counting and stats command as I originally had it is an option; I wanted it to be easier to "install" and therefore combined them not thinking it'd be this close.

**Status**

### DIFF
--- a/src/fun/counting_v2.go.tmpl
+++ b/src/fun/counting_v2.go.tmpl
@@ -44,7 +44,7 @@
 {{end}}
 
 {{if ($re2:=reFindAllSubmatches (print `(?i)\A(?:-|<@!?` .BotUser.ID `>)\s*C(?:ount)?(?:ing)?Stat(?:istic)?s?\s*(?:(?:<@!?)?(\d{17,})|(l(?:eader)?b))?`) .Cmd)}}
-	{{if and (index $re2 0 1|eq "") (index $re2 0 2|eq "")}} {{/* general stats */}}
+	{{if not (or (index $re2 0 1) (index $re2 0 2))}} {{/* general stats */}}
 		{{sendMessage nil (cembed 
 			"author" (sdict 
 				"icon_url" (.Guild.IconURL "512") 
--- a/src/fun/counting_v2.go.tmpl
+++ b/src/fun/counting_v2.go.tmpl
@@ -136,7 +136,7 @@
 		{{- $exp =printf "%s%s" $exp .}}
 		{{- if reFind `\d+` .}}
 			{{- if $lastNumeral}}
-				{{$invalid =true}}{{- break}}
+				{{- $invalid =true}}{{break}}
 			{{- end}}
 			{{- $lastNumeral =true}}
 		{{- else}}
--- a/src/fun/counting_v2.go.tmpl
+++ b/src/fun/counting_v2.go.tmpl
@@ -133,7 +133,7 @@
 {{$exp:=""}}{{$lastNumeral:=false}}
 {{range $re}}
 	{{- $roman:=reFindAll `[IVXLCDM]` .}}
-	{{- if eq (len $roman) 0}}
+	{{- if not $roman}}
 		{{- $exp =printf "%s%s" $exp .}}
 		{{- if reFind `\d+` .}}
 			{{- if $lastNumeral}}
--- a/src/fun/counting_v2.go.tmpl
+++ b/src/fun/counting_v2.go.tmpl
@@ -121,7 +121,7 @@
 {{end}}
 
 {{$invalid:=false}}
-{{$re:=reFindAll `\b([%e]|abs|pi)(\b|$)|[+\-*/^π()]|a?sin|a?cos|a?tan|a?cot|a?sec|a?csc|sqrt|l(o?g|n)|\d+|[MDCLXVI]M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})` .Cmd}}
+{{$re:=reFindAll `\b([%e]|abs|pi)(\b|$)|[+\-*/^π()]|a?(sin|cos|tan|cot|sec|csc)|sqrt|l(o?g|n)|\d+|[MDCLXVI]M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})` .Cmd}}
 {{if not (joinStr "" $re)}}
 	{{return}}
 {{end}}
--- a/src/fun/counting_v2.go.tmpl
+++ b/src/fun/counting_v2.go.tmpl
@@ -20,23 +20,23 @@
 {{ $leaderboardLength := 10 }} {{/* how many members to show on leaderboard; max of 100 */}}
 {{/* end of configurable values */}}
 
-{{$db:=or 
-	(dbGet 0 "counting").Value 
+{{$db:=or
+	(dbGet 0 "counting").Value
 	(sdict "last" (sdict "user" .BotUser.ID "msg" 0) "next" 1 "highscore" (sdict "user" .BotUser.ID "num" 0 "time" currentTime) "saves" $saves)
 }}
 
 {{with .ExecData}}
 	{{$msg:=getMessage nil .}}
 	{{if not $msg}} {{/* check if message was deleted */}}
-		{{sendMessage nil (cembed 
-			"description" (printf "<@%d> deleted their expression which was correct!\nThe next number is %d" 
+		{{sendMessage nil (cembed
+			"description" (printf "<@%d> deleted their expression which was correct!\nThe next number is %d"
 				$db.last.user $db.next
-			) 
+			)
 			"color" 30654
 		)}}
 	{{else if $msg.EditedTimestamp}} {{/* check if message was edited */}}
-		{{sendMessage nil (cembed 
-			"description" (printf "<@%d> edited their expression, be careful" $db.last.user) 
+		{{sendMessage nil (cembed
+			"description" (printf "<@%d> edited their expression, be careful" $db.last.user)
 			"color" 30654
 		)}}
 	{{end}}
@@ -45,18 +45,18 @@
 
 {{if ($re2:=reFindAllSubmatches (print `(?i)\A(?:-|<@!?` .BotUser.ID `>)\s*C(?:ount)?(?:ing)?Stat(?:istic)?s?\s*(?:(?:<@!?)?(\d{17,})|(l(?:eader)?b))?`) .Cmd)}}
 	{{if not (or (index $re2 0 1) (index $re2 0 2))}} {{/* general stats */}}
-		{{sendMessage nil (cembed 
-			"author" (sdict 
-				"icon_url" (.Guild.IconURL "512") 
+		{{sendMessage nil (cembed
+			"author" (sdict
+				"icon_url" (.Guild.IconURL "512")
 				"name" "🔢 Counting Statistics"
-			) 
-			"description" (printf "⌚ __Current Score__: %d\n🏅 __High Score__: %d on %v by `%s` (%d)\n⏮️ __Last Counter__: `%s` (%d)\n💾 __Saves Remaining__: %d" 
-				(sub $db.next 1) 
-				$db.highscore.num (formatTime $db.highscore.time "01/02") ($user:=userArg $db.highscore.user).Username $user.ID 
-				($user =userArg $db.last.user).Username $user.ID 
+			)
+			"description" (printf "⌚ __Current Score__: %d\n🏅 __High Score__: %d on %v by `%s` (%d)\n⏮️ __Last Counter__: `%s` (%d)\n💾 __Saves Remaining__: %d"
+				(sub $db.next 1)
+				$db.highscore.num (formatTime $db.highscore.time "01/02") ($user:=userArg $db.highscore.user).Username $user.ID
+				($user =userArg $db.last.user).Username $user.ID
 				$db.saves
-			) 
-			"footer" (sdict "text" "Use this command: -CStats [User/\"Leaderboard\"]") 
+			)
+			"footer" (sdict "text" "Use this command: -CStats [User/\"Leaderboard\"]")
 			"color" 30654
 		)}}
 		{{return}}
@@ -65,23 +65,23 @@
 	{{if ($user:=index $re2 0 1|toInt|userArg)}} {{/* user stats */}}
 		{{$userCount:=dbGet $user.ID "counting"}}
 		{{if not $userCount}}
-			{{sendMessage nil (cembed 
-				"title" (print "No available stats") 
-				"description" (printf "`%s` has yet to count ☹️\nMaybe give them a heads-up to come join?" 
+			{{sendMessage nil (cembed
+				"title" (print "No available stats")
+				"description" (printf "`%s` has yet to count ☹️\nMaybe give them a heads-up to come join?"
 					$user.Username
-				) 
-				"footer" (sdict "text" "Use this command: -CStats [User: @/ID]") 
+				)
+				"footer" (sdict "text" "Use this command: -CStats [User: @/ID]")
 				"color" 16711680
 			)}}
 		{{else}}
 			{{$userCount:=toInt $userCount.Value}}
 			{{$userCorrect:=(dbGet $user.ID "countingCorrect").Value}}
-			{{sendMessage nil (cembed 
-				"title" (printf "🔢 %s's Counting Statistics" $user.Username) 
-				"description" (printf "%s has counted **%d times**\n**%d** have been correct\nThis makes their average **%.2f%%**" 
+			{{sendMessage nil (cembed
+				"title" (printf "🔢 %s's Counting Statistics" $user.Username)
+				"description" (printf "%s has counted **%d times**\n**%d** have been correct\nThis makes their average **%.2f%%**"
 					$user.Mention $userCount (toInt $userCorrect) (div $userCorrect $userCount|mult 100.0)
-				) 
-				"footer" (sdict "text" "Use this command: -CStats [User: @/ID]") 
+				)
+				"footer" (sdict "text" "Use this command: -CStats [User: @/ID]")
 				"color" 30654
 			)}}
 		{{end}}
@@ -91,20 +91,20 @@
 		{{range $i,$v:=dbTopEntries "countingCorrect" $leaderboardLength 0}}
 			{{- $desc =printf "%s\n#%-3d %4d - %-4s" $desc (add $i 1) (toInt $v.Value) (or (userArg $v.UserID) (str $v.UserID)) -}}
 		{{end}}
-		{{sendMessage nil (cembed 
-			"author" (sdict 
-				"icon_url" (.Guild.IconURL "512") 
+		{{sendMessage nil (cembed
+			"author" (sdict
+				"icon_url" (.Guild.IconURL "512")
 				"name" "Counting Leaderboard"
-			) 
-			"description" (printf "```Pos    ✅  User%s```" $desc) 
-			"footer" (sdict "text" "Use this command: -CStats Leaderboard") 
+			)
+			"description" (printf "```Pos    ✅  User%s```" $desc)
+			"footer" (sdict "text" "Use this command: -CStats Leaderboard")
 			"color" 30654
 		)}}
 
 	{{else}}
-		{{sendMessage nil (cembed 
-			"title" "Invalid Syntax" 
-			"description" "For server counting statistics: `-CStats`\nFor a members' statistics: `-CStats <User: @/ID>`\nFor server leaderboard: `-CStats Leaderboard`" 
+		{{sendMessage nil (cembed
+			"title" "Invalid Syntax"
+			"description" "For server counting statistics: `-CStats`\nFor a members' statistics: `-CStats <User: @/ID>`\nFor server leaderboard: `-CStats Leaderboard`"
 			"color" 16744192
 		)}}
 	{{end}}
@@ -112,8 +112,8 @@
 {{end}}
 
 {{if and (eq $db.last.user .User.ID) (not $countTwice)}}
-	{{sendMessage nil (cembed 
-		"description" (printf "You can't count twice in a row 🥲\nThe next number is %d" $db.next) 
+	{{sendMessage nil (cembed
+		"description" (printf "You can't count twice in a row 🥲\nThe next number is %d" $db.next)
 		"color" 16744192
 	)}}
 	{{return}}
@@ -177,8 +177,8 @@
 {{catch}}{{$invalid =true}}{{end}}
 
 {{if $invalid}}
-	{{sendMessage nil (cembed 
-		"description" (printf "`%s` sent an invalid expression, `%s`\nNext number is `%d`" .User.Username .Cmd $db.next) 
+	{{sendMessage nil (cembed
+		"description" (printf "`%s` sent an invalid expression, `%s`\nNext number is `%d`" .User.Username .Cmd $db.next)
 		"color" 16711680
 	)}}
 	{{return}}
@@ -206,8 +206,8 @@
 		{{takeRoleID $db.last.user .}}
 		{{addRoleID .}}
 	{{end}}
-	{{$db.Set "last" (sdict 
-		"user" .User.ID 
+	{{$db.Set "last" (sdict
+		"user" .User.ID
 		"msg" .Message.ID
 	)}}
 
@@ -242,14 +242,14 @@
 		{{end}}
 		{{$db.Set "last" (sdict "user" .User.ID "msg" .Message.ID)}}
 		{{dbSet 0 "counting" $db}}
-		{{sendMessage nil (cembed 
-			"description" (printf "%s sent an incorrect number of %d\n**But**, the count was saved! %d saves remaining\nThe next number is %d" .User.Username $number $db.saves $db.next) 
+		{{sendMessage nil (cembed
+			"description" (printf "%s sent an incorrect number of %d\n**But**, the count was saved! %d saves remaining\nThe next number is %d" .User.Username $number $db.saves $db.next)
 			"color" 16744192
 		)}}
 
 	{{else}} {{/* reset count */}}
-		{{sendMessage nil (cembed 
-			"description" (printf "%s sent an incorrect number of %d\nCorrect number was %d\nStart over at 1 🙃" .User.Mention $number $db.next) 
+		{{sendMessage nil (cembed
+			"description" (printf "%s sent an incorrect number of %d\nCorrect number was %d\nStart over at 1 🙃" .User.Mention $number $db.next)
 			"color" 16711680
 		)}}
 		{{$db.Set "last" (sdict "user" .BotUser.ID "msg" 0)}}
--- a/src/fun/counting_v2.go.tmpl
+++ b/src/fun/counting_v2.go.tmpl
@@ -87,10 +87,9 @@
 		{{end}}
 
 	{{else if eq (index $re2 0 2|lower) "leaderb" "lb"}} {{/* leaderboard */}}
-		{{$desc:=""}}{{$pos:=1}}
-		{{range dbTopEntries "countingCorrect" $leaderboardLength 0}}
-			{{- $desc =printf "%s\n#%-3d %4d - %-4s" $desc $pos (toInt .Value) (or (userArg .UserID) (str .UserID))}}
-			{{- $pos =add $pos 1 -}}
+		{{$desc:=""}}
+		{{range $i,$v:=dbTopEntries "countingCorrect" $leaderboardLength 0}}
+			{{- $desc =printf "%s\n#%-3d %4d - %-4s" $desc (add $i 1) (toInt $v.Value) (or (userArg $v.UserID) (str $v.UserID)) -}}
 		{{end}}
 		{{sendMessage nil (cembed 
 			"author" (sdict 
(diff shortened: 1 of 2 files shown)


src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+	(dbGet 0 "counting").Value 
+	(sdict "last" (sdict "user" .BotUser.ID "msg" 0) "next" 1 "highscore" (sdict "user" .BotUser.ID "num" 0 "time" currentTime) "saves" $saves)
+}}
+
+{{with .ExecData}}
+	{{$msg:=getMessage nil .}}
+	{{if not $msg}} {{/* check if message was deleted */}}
+		{{sendMessage nil (cembed 
+			"description" (printf "%s deleted their expression which was correct!\nThe next number is %d" 
+				(userArg $db.last.user).Mention $db.next
```
jo3-l, line 33: You can save a `userArg` call with `<@%s> deleted ...` and passing in the user ID directly.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+			) 
+			"color" 30654
+		)}}
+	{{else if $msg.EditedTimestamp}} {{/* check if message was edited */}}
+		{{sendMessage nil (cembed 
+			"description" (printf "%s edited their expression, be careful" (userArg $db.last.user).Mention)
```
jo3-l, line 39: ditto

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+{{if and (eq $db.last.user .User.ID) (not $countTwice)}}
+	{{sendMessage nil (cembed 
+		"description" (printf "You can't count twice in a row 🥲\nThe next number is %d" $db.next) 
+		"color" 16744192
+	)}}
+	{{return}}
+{{end}}
+
+{{$invalid:=false}}
+{{$re:=reFindAll `\+|\-|\*|\/|\^|\%|a?sin|a?cos|a?tan|a?cot|a?sec|a?csc|sqrt|l(o?g|n)|abs|e|pi|π|\(|\)|\d+|M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})` .Cmd}}
```
jo3-l, line 124: You can fold the single characters into a character class: `[+-*/^%]`.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+		{{try}}
+			{{addReactions $correctEmoji}}
+			{{if and $reactionDelete $db.last.msg}}
+				{{deleteMessageReaction nil $db.last.msg .BotUser.ID $correctEmoji}}
+			{{end}}
+			{{if eq (mod $number 100|toInt) 0}}{{addReactions "💯"}}{{end}}
+		{{catch}}
+			{{with getRole $incorrectRID}}
+				{{addRoleID .ID}}{{(toDuration "1d").Seconds|toInt|removeRoleID .ID}}
+			{{end}}
```
jo3-l, line 199: You can use `addRole` with the role object directly, and likewise with `removeRole`.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+			{{sendMessage $errorCID (printf $errFormat . $.Message.Link $.User.Username $.User.ID)}}
+		{{end}}
+	{{end}}
+	{{with getRole $correctRID}}
+		{{takeRoleID $db.last.user .ID}}
+		{{addRoleID .ID}}
```
jo3-l, line 205: ditto

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+		{{end}}
+		{{$db.Set "highscore" (sdict "user" .User.ID "num" $number "time" currentTime)}}
+	{{end}}
+	{{dbSet 0 "counting" $db}}
+	{{execCC .CCID nil 15 .Message.ID}}
+		
+{{else}} {{/* incorrect */}}
+	{{$db.Set "saves" (sub $db.saves 1)}}
+	{{with getRole $correctRID}}
+		{{takeRoleID $db.last.user .ID}}
```
jo3-l, line 226: ditto

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+	{{end}}
+	{{with getRole $incorrectRID}}
+		{{addRoleID .ID}}{{(toDuration "3d").Seconds|toInt|removeRoleID .ID}}
```
jo3-l, line 229: ditto

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+	{{end}}
+	{{if ge $db.saves 0}} {{/* progress saved */}}
+		{{if $reactions}}
+			{{try}}{{addReactions $warningEmoji}}
+			{{catch}}
+				{{with getRole $incorrectRID}}
+					{{addRoleID .ID}}{{(toDuration "1d").Seconds|toInt|removeRoleID .ID}}
```
jo3-l, line 236: ditto

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,268 @@
...
+		{{$db.Set "last" (sdict "user" .BotUser.ID "msg" 0)}}
+		{{$db.Set "next" 1}}
+		{{$db.Set "saves" $saves}}
+		{{dbSet 0 "counting" $db}}
+		{{$s:=dbIncr .User.ID "counting" 1}}
+		{{if $reactions}}
+			{{try}}{{addReactions $incorrectEmoji}}
+			{{catch}}
+				{{with getRole $incorrectRID}}
+					{{addRoleID .ID}}{{(toDuration "1d").Seconds|toInt|removeRoleID .ID}}
```
jo3-l, line 262: ditto

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Counting System
+---
+
+This command controls the counting game and the statistics command.
```
jo3-l, line 5: A brief comment about what additional features your command offers over the previous counting command would be nice.

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Trigger
+
+**Type:** `Regex`<br />
+**Trigger:** `.+`
```
jo3-l, line 10: Note that `.+` does not match text split over multiple lines. I have not examined your code thoroughly but it seems like this could be a concern; for instance, would
```
1
<random text here>
```
be marked as invalid, assuming `1` is the next number?

Note also that this does not trigger on image-only posts, which might be abused? Not sure.

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,74 @@
...
+  ID of role for most recent correct counting user.
+
+- `$incorrectRID`<br />
+  ID of role for incorrect counting users.
+
+- `$errorCID`<br />
+  ID of channel for any error output.
+
+- `$saves`<br />
+  Amount of saves from incorrect counts before the count resets.
```
jo3-l, line 46: It's not clear what "saves from incorrect counts" means. Elaborate?

jo3-l (changes requested):

src/fun/counting_v2.go.tmpl
H1nr1: Out of curiosity, what benefit does this have; just one less step to get the ID value?

jo3-l: I assumed the length of the code was a minor concern given the compressed style, and `addRole .` is shorter without compromising too much on clarity.

website/docs/fun/counting_v2.md
H1nr1: Your example is valid and correct given that the number or expression on the first line evaluates to the next number.
I'm unsure how image-only posts could be used to abuse the system. My intent in many of my commands has been to allow "outside conversation" within the channel while not disrupting the game at hand. To do this, but also allow for such a wide range of input, I chose this trigger.

website/docs/fun/counting_v2.md
H1nr1: The amount of times an incorrect number or expression can be made before the count resets.
If $saves is 1, the first incorrect number sent will get a pass, but the second will reset the count to 1.

website/docs/fun/counting_v2.md
jo3-l: Fair enough. That is a significant difference from most counting systems I have seen in the wild, so I think it is worth mentioning explicitly in the description of this command (that unrelated content is permitted.)

website/docs/fun/counting_v2.md
jo3-l: Thanks, but I meant that you should clarify the documentation.

website/docs/fun/counting_v2.md
H1nr1: As messages that do include text content will trigger the command and an output if all characters do not match the available numerals, operators, or functions, maybe my claim for "outside conversation" is invalid for this command.
For example, if a user were to post "1+1 equals 2", this would be taken as invalid and a message would be output to show that.
So while messages without text will be ignored/permitted, messages with text while maybe unrelated, will not be entirely ignored. They will not affect the system or modify internal values.
So where I am now is how to word this in the description, or if it's even worth including.

jo3-l: We should certainly document that the command does not reject invalid inputs by deleting messages, since many other counting systems do so and users might mistakenly assume that this command does the same. Including additional detail beyond that is your choice; I do not have a strong opinion on documenting that.

H1nr1:
I have updated both files to include your changes in how I interpret them. The command file was straightforward, but the documentation may could use better wording; I'm unsure how it reads from another perspective.

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Counting System
+---
+
+This command controls the counting game and the statistics command.
```
jo3-l, line 5:
```suggestion
This command implements a counting game with detailed statistics and a leaderboard.
```

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,80 @@
...
+
+This version of a counting command allows for the usage of base 10 and roman numerals, optionally joined by the usage of math functions and/or operators to count from 1, incrementing by 1, until failure.
```
jo3-l, line 7:
```suggestion
A unique feature of this command is that, besides decimal numbers, it also accepts roman numerals and mathematical expressions as valid inputs. So if the next number is 2, `II` and `2*sin(pi/2)` would be considered correct as well.
```
I've not checked whether `2*sin(pi/2)` actually works with your code; replace with an appropriate example if not.

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,80 @@
...
+
+Also included are statistics for the server and each user for how many times they have counted and how many times they've been correct.
```
jo3-l, line 9: You can drop this sentence after changing the first line of the documentation as suggested earlier.

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,80 @@
...
+
+Messages without text are ignored by the system, effectively allowing unrelated content. However, messages including text which cannot be translated into a math expression will be quoted as invalid, but not deleted.
```
jo3-l, line 11: I think this should be in a `::note` callout for visibility. Concerning the content, I suggest
```suggestion
Unlike some other counting systems, this command does not delete invalid numbers to allow counting and other conversation to take place concurrently. Messages that include text which do not correspond to valid numbers will, however, trigger a warning message.
```

A clarifying question: is it the case that completely unrelated messages that do not contain anything resembling a number or a mathematical expression (so something like `Hello there!`) trigger the invalid input warning as well? If so, I would think this would disrupt conversation to the extent that you might as well disallow invalid inputs completely and delete them as most counting games do. But if you feel strongly that this command is fine as is, I am happy to drop this point.

jo3-l (changes requested):
Thanks, your latest changes look good. Some more nits about the documentation:

website/docs/fun/counting_v2.md
H1nr1: Just checked, it does. The built-in calc command handles the calculation, the CC just converts values into an acceptable expression and invalidates if necessary.

website/docs/fun/counting_v2.md
H1nr1: Yes, messages such as `Hello there!` will trigger the warning message. I agree with you though, and will work on allowing messages as such without warning for them.

H1nr1: Messed around with the expression regex and trying to find logic that would determine the message is not meant to be an expression. Issue is the regex, due to the part which captures roman numerals, also captures empty strings, which I'd prefer it not but cannot figure out how to require at least one roman numeral (I, V, X, L, C, D, M) for it to be captured as a roman numeral. Another issue is, using the logic `if eq (joinStr "" $re) ""`, if the message includes the common letter e, or another character related to math, it evaluates to false, therefore disqualifying that method.

H1nr1: Update: As of the latest commit - which makes a change to the regex which parses the expression and adds a conditional statement to make sure an expression was parsed - I believe this command to act fully as intended. 

Messages which are not meant to be expressions are to be ignored, message which are attempted expressions but have invalid syntax or include unaccepted characters/"functions" are to be given a warning message, and valid expressions are to be accepted and evaluated towards the correct next number.

H1nr1:
Sorry, I know you're busy with docs v2 and I'm sure much more, but could I get a status update on this? Thanks!

jo3-l (approved):
This looks OK to me. I would like @l-zeuch to take a second look at the documentation here if he has time—the code is probably fine as is; major modifications are probably unfeasible given the character limit anyway. If not, we can merge this sometime in the next few days.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+{{if and (eq $db.last.user .User.ID) (not $countTwice)}}
+	{{sendMessage nil (cembed 
+		"description" (printf "You can't count twice in a row 🥲\nThe next number is %d" $db.next) 
+		"color" 16744192
+	)}}
+	{{return}}
+{{end}}
+
+{{$invalid:=false}}
+{{$re:=reFindAll `\b([%e]|abs|pi)(\b|$)|[+\-*/^π()]|a?sin|a?cos|a?tan|a?cot|a?sec|a?csc|sqrt|l(o?g|n)|\d+|[MDCLXVI]M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})` .Cmd}}`{{json $re}}`
```
galen8183, line 124: Looks like you accidentally left `` `{{json $re}}` `` in here.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+				"color" 30654
+			)}}
+		{{end}}
+
+	{{else if eq (index $re2 0 2|lower) "leaderb" "lb"}} {{/* leaderboard */}}
+		{{$desc:=""}}{{$pos:=1}}
+		{{range dbTopEntries "countingCorrect" $leaderboardLength 0}}
+			{{- $desc =printf "%s\n#%-3d %4d - %-4s" $desc $pos (toInt .Value) (or (userArg .UserID) (str .UserID))}}
+			{{- $pos =add $pos 1 -}}
+		{{end}}
```
galen8183, line 94:
```suggestion
		{{$desc:=""}}
		{{range $i,$v:=dbTopEntries "countingCorrect" $leaderboardLength 0}}
			{{- $desc =printf "%s\n#%-3d %4d - %-4s" $desc (add $i 1) (toInt $v.Value) (or $v.User.String (str $v.UserID))}}
		{{end}}
```
This uses the iteration index instead of manually incrementing `$pos` and saves a `userArg` call.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+		{{sendMessage nil (cembed 
+			"description" (printf "<@%d> edited their expression, be careful" $db.last.user) 
+			"color" 30654
+		)}}
+	{{end}}
+	{{return}}
+{{end}}
+
+{{if ($re2:=reFindAllSubmatches (print `(?i)\A(?:-|<@!?` .BotUser.ID `>)\s*C(?:ount)?(?:ing)?Stat(?:istic)?s?\s*(?:(?:<@!?)?(\d{17,})|(l(?:eader)?b))?`) .Cmd)}}
+	{{if and (index $re2 0 1|eq "") (index $re2 0 2|eq "")}} {{/* general stats */}}
```
galen8183, line 47:
```suggestion
	{{if not (or (index $re2 0 1) (index $re2 0 2))}} {{/* general stats */}}
```
I think this logic makes more sense here.

src/fun/counting_v2.go.tmpl
H1nr1: Whoops... not sure how I missed that, I've copied the code from here multiple times to make sure it's correct.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+	{{sendMessage nil (cembed 
+		"description" (printf "You can't count twice in a row 🥲\nThe next number is %d" $db.next) 
+		"color" 16744192
+	)}}
+	{{return}}
+{{end}}
+
+{{$invalid:=false}}
+{{$re:=reFindAll `\b([%e]|abs|pi)(\b|$)|[+\-*/^π()]|a?sin|a?cos|a?tan|a?cot|a?sec|a?csc|sqrt|l(o?g|n)|\d+|[MDCLXVI]M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})` .Cmd}}`{{json $re}}`
+{{if eq (joinStr "" $re) ""}}
```
galen8183, line 125:
```suggestion
{{if not (joinStr "" $re)}}
```
You can use truthiness here.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+{{end}}
+{{if ne .Cmd (joinStr "" $re)}}
+	{{$invalid =true}}
+{{end}}
+
+{{$convertRN:=sdict "I" 1 "V" 5 "X" 10 "L" 50 "C" 100 "D" 500 "M" 1000}}
+{{$exp:=""}}{{$lastNumeral:=false}}
+{{range $re}}
+	{{- $roman:=reFindAll `I|V|X|L|C|D|M` .}}
+	{{- if eq (len $roman) 0}}
```
galen8183, line 136:
```suggestion
	{{- if not $roman}}
```
Truthiness here too.

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+			{{- $lastNumeral =true}}
+		{{- else}}
+			{{- $lastNumeral =false}}
+		{{- end}}
+		{{- continue}}
+	{{- end}}
+	{{- if $lastNumeral}}
+		{{$invalid =true}}{{- break}}
+	{{- end}}
+	{{- $values:=cslice}}{{- $value:=0}}{{- $skip:=false}}
```
galen8183, line 151:
```suggestion
	{{- $values:=cslice}}{{$value:=0}}{{$skip:=false}}
```

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+				{{$invalid =true}}{{- break}}
+			{{- end}}
+			{{- $lastNumeral =true}}
+		{{- else}}
+			{{- $lastNumeral =false}}
+		{{- end}}
+		{{- continue}}
+	{{- end}}
+	{{- if $lastNumeral}}
+		{{$invalid =true}}{{- break}}
```
galen8183, line 149:
```suggestion
		{{- $invalid =true}}{{break}}
```

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+
+{{$convertRN:=sdict "I" 1 "V" 5 "X" 10 "L" 50 "C" 100 "D" 500 "M" 1000}}
+{{$exp:=""}}{{$lastNumeral:=false}}
+{{range $re}}
+	{{- $roman:=reFindAll `I|V|X|L|C|D|M` .}}
+	{{- if eq (len $roman) 0}}
+		{{- $exp =printf "%s%s" $exp .}}
+		{{- if reFind `\d+` .}}
+			{{- if $lastNumeral}}
+				{{$invalid =true}}{{- break}}
```
galen8183, line 140:
```suggestion
				{{- $invalid =true}}{{break}}
```
The existing trim marker does nothing.

galen8183 (commented):
Some minor style changes.
There are trailing spaces on many lines, please remove those if they don't provide any purpose to the command response.

src/fun/counting_v2.go.tmpl
H1nr1: While it may be more formal to use a non-capturing group for the trig functions, is it necessary given YAG does not keep track of capture groups?

src/fun/counting_v2.go.tmpl
```diff
@@ -0,0 +1,271 @@
...
+{{if and (eq $db.last.user .User.ID) (not $countTwice)}}
+	{{sendMessage nil (cembed 
+		"description" (printf "You can't count twice in a row 🥲\nThe next number is %d" $db.next) 
+		"color" 16744192
+	)}}
+	{{return}}
+{{end}}
+
+{{$invalid:=false}}
+{{$re:=reFindAll `\b([%e]|abs|pi)(\b|$)|[+\-*/^π()]|a?sin|a?cos|a?tan|a?cot|a?sec|a?csc|sqrt|l(o?g|n)|\d+|[MDCLXVI]M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})` .Cmd}}
```
galen8183, line 124:
```suggestion
{{$re:=reFindAll `\b([%e]|abs|pi)(\b|$)|[+\-*/^π()]|a?(?:sin|cos|tan|cot|sec|csc)|sqrt|l(o?g|n)|\d+|[MDCLXVI]M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})` .Cmd}}
```
Without looking too hard this can be slightly simplified.

H1nr1:
> Some minor style changes. There are trailing spaces on many lines, please remove those if they don't provide any purpose to the command response.

Do you mean before comments? I've never thought anything of it, but yes, that'd definitely be included in the response here when I didn't intend to.

src/fun/counting_v2.go.tmpl
galen8183: My bad, I just added another review comment since I didn't notice you had seen this already... you're welcome to remove that if you'd like, it's a personal style choice :)

galen8183:
> > Some minor style changes. There are trailing spaces on many lines, please remove those if they don't provide any purpose to the command response.
> 
> Do you mean before comments? I've never thought anything of it, but yes, that'd definitely be included in the response here when I didn't intend to.

It seems random to me, they're spaces at the very end of lines, not before comments.

src/fun/counting_v2.go.tmpl
H1nr1: I like the idea of using iteration index, however, the change to save the `userArg` call presents an issue. It seems that `or $v.User.String (str $v.UserID)` returns `#` for users not in the server. I'm not sure why this is; I didn't know entries made `.User.String` available as docs state "User object of type *discordgo.User* having only `.ID` field, .Mention is still usable with correct userID field entry."

src/fun/counting_v2.go.tmpl
H1nr1: Not needed because they're on the same line? That would be a new idea to me

H1nr1:
> > > Some minor style changes. There are trailing spaces on many lines, please remove those if they don't provide any purpose to the command response.
> > 
> > 
> > Do you mean before comments? I've never thought anything of it, but yes, that'd definitely be included in the response here when I didn't intend to.
> 
> It seems random to me, they're spaces at the very end of lines, not before comments. If your text editor has a regex find and replace you can search for ` $`

I see, I do this because if the code was to be inline, and not indented as such, spaces would be there. I understand the new lines pose as the whitespace, and therefore the spaces aren't needed. If still requested, I can remove them.
[example image](https://i.imgur.com/6Ec1mlv.png)

src/fun/counting_v2.go.tmpl
galen8183: My bad, I misremembered that it isn't populated, `userArg` is needed then.

src/fun/counting_v2.go.tmpl
galen8183: Trim markers only trim whitespace from the action they're added to and the last/next non-whitespace character.

galen8183:
It isn't inline, so why keep them? If you want to change them to a single line you can simply replace each newline and indentation with a space.

H1nr1:
I do it somewhat for if I want to make the code inline, it makes it easier to do so. I will remove the extra spaces.

As for the spaces before comments, should they be removed too?

galen8183:
`s/\n\t*/ ` is just as easy... the spaces before comments are okay

website/docs/fun/counting_v2.md
```diff
@@ -0,0 +1,82 @@
...
+title: Counting System
+---
+
+This command implements a counting game with detailed statistics and a leaderboard.
+
+A unique feature of this command is that, besides decimal numbers, it also accepts roman numerals and mathematical expressions as valid inputs. So if the next number is 2, `II` and `2*sin(pi/2)` would be considered correct as well.
+
+:::note
+
+Unlike some other counting systems, this command does not delete invalid numbers to allow counting and other conversation to take place concurrently. Messages that include text which do not correspond to valid numbers will, however, trigger a warning message.
```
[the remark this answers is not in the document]
H1nr1: I think your first suggestion reads somewhat incorrectly; it reads to me that if I were to throw a 1 somewhere in the message, it would be valid, even if the expression sent is `invalid1exp`. However, this does make me notice the usage of the word "numbers" here, which may make the user question if terms other than decimal numbers are accepted. I think it'd be better to use "terms" in replacement.

As for the question and suggestion on first line expressions, yes, only the first line of the message is parsed for an expression. I believe the previous paragraph was the result of a [previous discussion](https://github.com/yagpdb-cc/yagpdb-cc/pull/334#discussion_r1685176559). It could use some expanding to make it known only the first line is parsed and the rest is ignored.

I will not make any changes to this now so it can be discussed.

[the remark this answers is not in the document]
H1nr1:
```suggestion
Unlike some other counting systems, this command does not delete invalid expressions to allow counting and other conversation to take place concurrently. Expressions which include unaccepted terms or evaluate to an invalid expression will, however, trigger a warning message.
```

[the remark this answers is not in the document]
l-zeuch: > I think your first suggestion reads somewhat incorrectly; it reads to me that if I were to throw a 1 somewhere in the message, it would be valid, even if the expression sent is invalid1exp

That was not the point of the first suggestion. The point was to simplify this sentence, the second suggestion expanded on that by asking to note down the behaviour with regards to the first line of a message.


> I think it'd be better to use "terms" in replacement.

No, you already defined what counts as a valid number for the purposes of this custom command, why introduce another term? There's "number" and "valid number", I don't see how replacing that with "expression" makes things clearer.

Even still, your suggestion doesn't make this any clearer anyway, as you're not stating that this "expression" has to occur on the first line.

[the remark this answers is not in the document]
H1nr1: I understand the point of the first suggestion was for simplification. However, I am saying that from the rewrite, it does not read the same.

I think using "number" in the docs will not influence the users to use the command to it's potential, which is fine if they don't want to use roman numerals and/or math. But I think it scopes in on decimal numbers as the message to send, and not the result of an expression.

I did not include the second line as I wasn't confident in including it, and can be committed separately after this discussion. Overall, yes, I agree it should be noted only the first line is to be parsed.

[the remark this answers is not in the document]
l-zeuch: Sounds like we came to an agreement. Add these notes and then we should be good to go.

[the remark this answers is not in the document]
H1nr1: I've updated the note to be more direct in my opinion and added the first line suggestion.

l-zeuch (approved):

PR #345, opened 2024-11-20
Reply to `x.com` links with `fixupx.com` links

laforcem:

**Status**

- [ x I have [written documentation](https://github.com/yagpdb-cc/yagpdb-cc/blob/master/WRITING-DOCUMENTATION.md) for my changes, or there is no need to

### DIFF
--- a/src/utilities/fixupx.go.tmpl
+++ b/src/utilities/fixupx.go.tmpl
@@ -1,6 +1,8 @@
 {{/*
 	Replies with fixupx.com links to messages containing x.com links.
 
+  See <https://yagpdb-cc.github.io/utilities/fixupx> for more information.
+
 	Author: laforcem <https://github.com/laforcem>
 */}}
 
--- a/src/utilities/fixupx.go.tmpl
+++ b/src/utilities/fixupx.go.tmpl
@@ -1,0 +1,25 @@
+{{/*
+	Replies with fixupx.com links to messages containing x.com links.
+
+	Author: laforcem <https://github.com/laforcem>
+*/}}
+
+{{/* Find all x.com links in the message */}}
+{{ $links := reFindAll `https:\/\/x\.com(?:\/[^\s]*)?` .Message.Content }}
+{{ $fixedLinks := "" }}
+
+{{/* Loop through each found link, replace it, and append to the message */}}
+{{ range $link := $links }}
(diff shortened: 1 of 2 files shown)


website/docs/utilities/fixupx.md
```diff
@@ -0,0 +1,37 @@
...
+Check out this Tweet! https://x.com/rickastley/status/890617797956456448
+```
+
+## Configuration
+
+In the custom command control panel, configure the command as follows:
+
+**Trigger type:** `Contains`  
+**Trigger:** `https://x.com`
+
```
savage4618, line 28: Doesn't this contradict the above regex trigger instructions?

laforcem: I was under the impression that the Trigger section is for the code itself, and not necessarily for the configuration in the control panel. But I see my mistake now - will get that corrected!

website/docs/utilities/fixupx.md
```diff
@@ -0,0 +1,30 @@
+---
+title: FixupX Reminder
```
jo3-l, line 2: I wonder if a title like "Replace X (Twitter) links with FixupX" would make this more discoverable for folks searching. What do you think?

src/utilities/fixupx.go.tmpl
```diff
@@ -0,0 +1,27 @@
+{{/*
+	Replies with fixupx.com links to messages containing x.com links.
+
+  See <https://yagpdb-cc.github.io/utilities/fixupx> for more information.
```
jo3-l, line 4: If it's not too much trouble, would you mind reindenting this file to consistently use tabs instead of spaces? It's a little inconsistent as is.

jo3-l (commented):
Thanks, this looks nice. A couple minor nits:

website/docs/utilities/fixupx.md
SoggySaussages: I guess the question is what is the intent of this CC, a reminder with a fixer added on or a fixer with a reminder added on? I agree with Joe that the latter would be more appealing, but maybe I'm biased because that's what I'd prefer on my server. Imo if you have a system on your server appending fixupx links then it seems excessive to be pressuring your members to change their links manually before sending. Just clogs up the chat I feel. Have you considered making the reminder part of the CC a config value off by default? I like the attention given to make the plurality correct :)

PR #347, opened 2025-01-17
fun/coinflip: rewrite

H1nr1:

Initiate a draft PR for coinflip.

I intended to remove the nesting of if statements. I have realized after a deeper review, that many things can and should be updated, so much so that I would rather rewrite the command in my own fashion. I believe coinflip is a simple command and does not necessarily need a second version; but how could we preserve David's involvement if I were to rewrite it? I will include most features in my rewrite, but notably removing the need for a channel list and `scheduleUniqueCC`.

**Status**

- [X] I have read and followed the [contribution guide](https://github.com/yagpdb-cc/yagpdb-cc/blob/master/CONTRIBUTING.md)
- [X] I have [written documentation](https://github.com/yagpdb-cc/yagpdb-cc/blob/master/WRITING-DOCUMENTATION.md) for my changes, or there is no need to

### DIFF
--- a/src/fun/coinflip.go.tmpl
+++ b/src/fun/coinflip.go.tmpl
@@ -92,8 +92,8 @@
 	{{ end }}
 
 	{{ $balance := dbIncr .User.ID $DB_KEY $bet | toInt }}
-	{{ $embed.Set "description" (printf "**----------**\nYou chose: **%s** | The coin rolled: **%s**.\nYou %s **%d** credits.\nNow you have **%d** credits.\n**----------**" $choice $result.side $result.WL $result.diff $balance) }}
-	{{ dbSetExpire .User.ID "cd_coinflip" true $COOLDOWN }}
-	{{ dbDel .User.ID "isrolling_cf" }}
+	{{ $embed.Set "description" (printf "**----------**\nYou chose **%s** | The coin landed on **%s**.\nYou %s **%d** %s.\nNow you have **%d** %s.\n**----------**" $choice $result.side $result.WL $result.diff (lower $DB_KEY) $balance (lower $DB_KEY)) }}
+	{{ dbSetExpire .User.ID "CF_cooldown" true $COOLDOWN }}
+	{{ dbDel .User.ID "CF_isFlipping" }}
 	{{ editMessage $channel $m (cembed $embed) }}
 {{ end }}
--- a/src/fun/coinflip.go.tmpl
+++ b/src/fun/coinflip.go.tmpl
@@ -43,12 +43,14 @@
 		{{ $err = printf "Maximum bet can only be **%d**!" $MAX_BET }}
 	{{ end }}
 
-	{{ if dbGet .User.ID "cd_coinflip" }}
-		{{ $err = print "You need to wait **" (humanizeDurationSeconds ((dbGet .User.ID "cd_coinflip").ExpiresAt.Sub currentTime)) "** to use this command again." }}
+	{{ $cd := dbGet .User.ID "cd_coinflip" }}
+	{{ if $cd }}
+		{{ $err = print "You need to wait **" ($cd.ExpiresAt.Sub currentTime | humanizeDurationSeconds) "** to use this command again." }}
 	{{ end }}
 
-	{{ if lt (toInt (dbGet .User.ID $DB_KEY).Value) $bet }}
-		{{ $err = printf "You dont have **%d** credits.\nYou only have **%d** credits." $bet (toInt (dbGet .User.ID $DB_KEY).Value) }}
+	{{ $userBalance := (dbGet .User.ID $DB_KEY).Value | toInt }}
+	{{ if lt $userBalance $bet }}
+		{{ $err = printf "You dont have **%d** credits.\nYou only have **%d** credits." $bet $userBalance }}
 	{{ end }}
 
 	{{ if dbGet .User.ID "isrolling_cf" }}
@@ -64,7 +66,7 @@
 		{{ return }}
 	{{ end }}
 
-	{{/* roll may proceed */}}
+	{{/* flip may proceed */}}
 	{{ dbSet .User.ID "isrolling_cf" true }}
 	{{ $embed.Set "image" (sdict "url" $img) }}
 	{{ $embed.Set "color" 16765696 }}
--- a/src/fun/coinflip.go.tmpl
+++ b/src/fun/coinflip.go.tmpl
@@ -3,118 +3,117 @@
 	See <https://yagpdb-cc.github.io/fun/coinflip> for more information.
 
 	Author: DaviiD1337 <https://github.com/DaviiD1337>
+	Co-Author: H1nr1 <https://github.com/H1nr1>
 */}}
 
 {{/* Configuration variables start */}}
 {{ $channel := .Channel.ID }}
-{{ $minbet := 100 }}
-{{ $maxbet := 500 }}
+{{ $minBet := 100 }}
+{{ $maxBet := 500 }}
 {{ $db := "CREDITS" }}
-{{ $cooldown := 5 }} {{/* cannot be 0 */}}
-{{ $channels := cslice
-  752497563298562118
-  722554898612355093
-  744670631601111222
-  782711001333235734
-  783021217279246356
- }}
+{{ $cooldown := 5 }}
 {{/* Configuration variables end */}}
 
-{{ range $channels }}
-	{{ if not (getChannel .) }}
-		{{ print "Error: Configuration is not set correct. Channel with ID  `" . "` not found. (`$channels`)" }}
-		{{ deleteTrigger 5 }}
-		{{ deleteResponse 5 }}
-		{{ return }}
-	{{ end }}
-{{ end }}
+{{ $err := false }}
+{{ $img := "https://cdn.discordapp.com/attachments/707661790443733022/782710794634264616/d74906d39a1964e7d07555e7601b06ad.gif" }}
+{{ $heads := "https://cdn.discordapp.com/attachments/707661790443733022/782713935341027358/cap.png" }}
+{{ $tails := "https://cdn.discordapp.com/attachments/707661790443733022/782713937979637760/pajura.png" }}
+{{ $embed := sdict "author" (sdict "name" .User.Username "icon_url" (.User.AvatarURL "256")) }}
 
-{{ $cp := "" }}{{ $m := 0 }}{{ $bet := "" }}{{ $err := false }}{{ $img := "https://cdn.discordapp.com/attachments/707661790443733022/782710794634264616/d74906d39a1964e7d07555e7601b06ad.gif" }}{{ $cap := "https://cdn.discordapp.com/attachments/707661790443733022/782713935341027358/cap.png" }}{{ $pajura := "https://cdn.discordapp.com/attachments/707661790443733022/782713937979637760/pajura.png" }}{{ $embed:= sdict }}{{ $fields := cslice }}{{ $embed.Set "author" (sdict "name" .User.Username "icon_url" (.User.AvatarURL "256"))}}
 {{ if not .ExecData }}
-	{{ if eq (len .CmdArgs) 2 }}
-		{{ with reFind `\A(?i)(heads|tails)\z` (index .CmdArgs 0) }}
-			{{ $cp = . | lower }}
-			{{ with reFindAll `\d+` (index $.CmdArgs 1) }}
-				{{ $bet = index . 0 | toInt }}
-				{{ if ge $bet $minbet }}
-					{{ if le $bet $maxbet }}
-						{{ if not (dbGet $.User.ID "cd_coinflip") }}
-							{{ if ge (toInt (dbGet $.User.ID $db).Value) $bet }}
-								{{ if not (dbGet $.User.ID "isrolling_cf") }}
-									{{ dbSet $.User.ID "isrolling_cf" true }}
-									{{ $embed.Set "image" (sdict "url" $img) }}
-									{{ $embed.Set "color" 16765696 }}
-									{{ $embed.Set "description" "**The coin is rolling... Wait...**" }}
-									{{ $m = sendMessageRetID $channel (cembed $embed) }}
-									{{ scheduleUniqueCC $.CCID (index (shuffle $channels) 0) 3 $.User.ID (sdict "m" $m "b" $bet "c" $cp )}}
-								{{ else }}
-									{{ $embed.Set "description" "You can use this command when the last coin stopped rolling!" }}
-									{{ $embed.Set "color" 16488706 }}
-									{{ $err = true }}
-								{{ end }}
-							{{ else }}
-								{{ $embed.Set "description" (printf "You dont have **%d** credits.\nYou only have **%d** credits." $bet (toInt (dbGet $.User.ID $db).Value)) }}
-								{{ $embed.Set "color" 16488706 }}
-								{{ $err = true }}
-							{{ end }}
-						{{ else }}
-							{{ $embed.Set "description" (print "You need to wait **" (humanizeDurationSeconds ((dbGet $.User.ID "cd_coinflip").ExpiresAt.Sub currentTime)) "** to use this command again.") }}
-							{{ $embed.Set "color" 16488706 }}
-							{{ $err = true }}
-						{{ end }}
-					{{ else }}
-						{{ $embed.Set "description" (printf "Maximum bet can only be **%d**!" $maxbet) }}
-						{{ $embed.Set "color" 16488706 }}
-						{{ $err = true }}
-					{{ end }}
-				{{ else }}
-					{{ $embed.Set "description" (printf "Minimum bet can only be **%d**!" $minbet) }}
-					{{ $embed.Set "color" 16488706 }}
-					{{ $err = true }}
-				{{ end }}
-			{{ else }}
-				{{ $embed.Set "description" (printf "`%s` is not a valid number." (index $.CmdArgs 1)) }}
-				{{ $embed.Set "color" 16488706 }}
-				{{ $err = true }}
-			{{ end }}
-		{{ else }}
-			{{ $embed.Set "description" "You can only choose from **heads** or **tails**." }}
-			{{ $embed.Set "color" 16488706 }}
-			{{ $err = true }}
-		{{ end }}
-	{{ else }}
+	{{ if not (eq (len .CmdArgs) 2) }}
 		{{ $embed.Set "description" (printf "Syntax: **%scoinflip <heads/tails> <money>**" .ServerPrefix) }}
 		{{ $embed.Set "color" 16488706 }}
 		{{ $err = true }}
 	{{ end }}
+
+	{{ $choice := reFind `\A(?i)(heads|tails)\z` (index .CmdArgs 0) }}
+	{{ if not $choice }}
+		{{ $embed.Set "description" "You can only choose from **heads** or **tails**." }}
+		{{ $embed.Set "color" 16488706 }}
+		{{ $err = true }}
+	{{ end }}
+
+	{{ $bet := reFind `\d+` (index .CmdArgs 1) | toInt }}
+	{{ if not $bet }}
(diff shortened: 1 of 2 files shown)


l-zeuch:
First of all, thanks for offering to update some older custom commands to use some of the "newer" features! It only makes sense that a curated collection should showcase reasonably good code.

As for keeping David's initial involvement, I believe it should be sufficient to keep the original author's attribution and just add your own below; both in the code and on the respective documentation page. I agree that a second version seems unnecessary; so long you keep feature parity with the current custom command, I see no reason why it should be separated out into a v2.

As far as I understand it, the `$channel` slice is to smear the load of `scheduleUniqueCC`--you're quite right that this is no longer necessary, as we have `.Guild.Channels` available, which we can use instead. Migration instructions should therefore be not necessary.

I suggest that you first start de-nesting the code in an initial commit, and then, once the code is already considerably cleaner, rework some of the more intricate paths; that way, code review becomes a whole lot easier on both ends.

DaviiD1337:
Right now I do not find enough time for a full rewrite, after the exams will look into it and try to re-write so the code look cleaner, or if someone has some spare time that would be great.

 Thanks for the PR.

H1nr1:
@l-zeuch Thank you for the in-depth reply.

I like the idea of de-nesting now, and refactoring later; I will proceed with that.

As for refactoring, I have a question relating to the `$channels` and `scheduleUniqueCC`. That is, would we rather use `sleep 3` or continue with `scheduleUniqueCC` and `.Guild.Channels`? I don't think a 3 second delay would cause too large of an issue with users using the command and could actually be used as a cooldown, without needing a cooldown via the database.

DaviiD1337:
I've decided not to use sleep because the game may get crowded, for example 5 players want to play the game in the same time, that will happen in a large server. For example for 5 players x 3 seconds sleep = 15 seconds for the last user of just waiting for other games to finish.

By using `scheduleUniqueCC` this can be avoided, but also let the server owner/admin change the duration of the coin flipping.

My part of view is by using `scheduleUniqueCC` and `.Guild.Channels` will do the best, but maybe you can add a new variable to the configuration so it uses `sleep` or `scheduleUniqueCC`, depending on the config, so admins can change to their preference.

H1nr1:
I think `scheduleUniqueCC` is definitely the preferred way to go for ease of use.

The only skepticism I have is if we were to use `.Guild.Channels`, are we okay with having `index .Guild.Channels (randInt (len .Guild.Channels))`? I'd store `.Guild.Channels` in a variable in the code, but is this "limit-evading" technique allowed on the public repo?

l-zeuch:
Setting a variable to the value of `.Guild.Channels` makes no difference; just stick with accessing context data directly.

Yes, `.Guild.Channels` was added knowing fully that `execCC` and friends would gain a huge buff--have fun with it.

H1nr1:
Ok, I thought it would be one less operation. Thanks!

l-zeuch:
> The updated version of the code currently works as Pedro initially implemented it.

Thanks, this already looks much cleaner and is easier to follow.

Some remarks about your suggestions follow.

> * Parsing: can be done in one `reFindAllSubmatches` statement

Personally I like your current implementation--as I understand our goal with this collection, we aim for code that is readable by everyone (superficially) familiar with custom commands. Having super clever constructs when we technically have the space available (i.e. not even close to 10_000 characters) only hurts that idea. Parsing individual parts as-needed seems the best way to go about this, but my opinion can surely be swayed in another direction. Let me know what you think about this sentiment.

> * Output: instead of `$embed.Set`, we could have an `execTemplate` call which sends the message and `return` after the call. In the `execTemplate` call, we can include the changes to `$embed`.

At the time I don't think this serves a great benefit to overall readability. Generally, we recommend sticking with a simple flow, and only use associated templates ("functions") if and only if it **significantly** shortens the entire code. Bear in mind that defining an associated template itself costs 100 ops, and each subsequent call costs 10 additional ops.
For now, just keep it as is; we can at a later point decide if it's worth it, so long it massively reduces code size.

> * Result: flip result could be simplified in that lines can be combined

I'm not entirely sure if I fully understand what you're saying here, could you perhaps provide a small snippet demonstrating this idea?

> 
> * Naming: some more variables and keys could be clarified for readability

I already see some changes to that regard; you're using `camelCase` (as we also do in the documentation) and reasonable variable names. Perhaps you could re-name configuration variables such that they use `UPPER_SNAKE_CASE`? It's what I do in my custom commands, at least. Other than that, this isn't a huge blocker and should also be rather put on the back-burner, similar to the idea of using associated templates.
 
> The error output I have multiple ideas for and am not sure what would be best. I think it is clear we can move the changing of the color to the `if $err` block as that is consistent. As for the description, it may be fine as is, or we could set `$err` to a string representing the error type, then get the error using an `sdict` in the `if $err` block.

We could just set `$err` to the error string and then `$embed.Set "description" $err` in our `if $err` block--is that consistent with what you're suggesting, or do you have a different approach in mind? Also, yeah, move the duplicate calls to `$embed.Set "color"` into a single line in the error handling block.

H1nr1:
> Parsing individual parts as-needed seems the best way to go about this

I agree after reading your comment. I was thinking the "clever" path would be best for the code, but failing to think about those using the code.
While keeping it easily readable, I believe we should use `.StrippedMsg` or `joinStr " " .CmdArgs`, instead of `index .CmdArgs #`. This would allow for the arguments to be in any order (there's only two here) which has shown helpful in other custom commands I've made when members cannot remember the command syntax. I did plan on allowing "h", "head", or "heads" for the heads choice, and likewise with tails, but I see the possibility of the command parsing heads if, for example, the user typed `-coinflip 100 hello tails`. So, my new suggested changes I am even uncertain with. Please let me know if we should go about this.

> - Result: flip result could be simplified in that lines can be combined

```diff
- {{ $chance := randInt 1 3 }}
- {{ $choosen := "" }}
+ {{ $result := (dict 
+ 	0 (sdict "side" "heads" "img" $heads)
+ 	1 (sdict "side" "tails" "img" $tails)
+ ).Get (randInt 2) }}

- {{ if eq $chance 1 }}
- 	{{ $choosen = "heads" }}
- 	{{ $embed.Set "image" (sdict "url" $heads) }}
- {{ else }}
- 	{{ $choosen = "tails" }}
- 	{{ $embed.Set "image" (sdict "url" $tails)}}
- {{ end }}
+ {{ $embed.Set "image" (sdict "url" $result.img) }}

- {{ if eq $choice $choosen }}
+ {{ if eq $choice $result.side }}
- 	{{ $k := dbIncr .User.ID $db $bet }}
- 	{{ $embed.Set "description" (printf "**----------**\nYou chose: **%s** | The coin rolled: **%s**.\nYou won **%d** credits.\nNow you have **%d** credits.\n**----------**" $choice $choosen (mult $bet 2) (toInt $k)) }}
+ 	{{ $result.Set "WL" "won" }}
+ 	{{ $result.Set "diff" (mult $bet 2) }}
	{{ $embed.Set "color" 40811 }}
{{ else }}
- 	{{ $k := dbIncr .User.ID $db (mult $bet -1) }}
- 	{{ $embed.Set "description" (printf "**----------**\nYou chose: **%s** | The coin rolled: **%s**.\nYou lost **%d** credits.\nNow you have **%d** credits.\n**----------**" $choice $choosen $bet (toInt $k)) }}
+ 	{{ $result.Set "WL" "lost" }}
+ 	{{ $result.Set "diff" $bet }}
+ 	{{ $bet = mult $bet -1 }}
	{{ $embed.Set "color" 16488706 }}
{{ end }}
+ {{ $balance := dbIncr .User.ID $db $bet | toInt }}
+ {{ $embed.Set "description" (printf "**----------**\nYou chose: **%s** | The coin rolled: **%s**.\nYou %s **%d** credits.\nNow you have **%d** credits.\n**----------**" $choice $result.side $result.WL $result.diff $balance) }}
```

After writing this, it seems like it would have its positives and negatives. While functions like `dbIncr` have two instances, only one of them would run. My update does not make a change to calls to the function, but combines the instances. It is arguable whether this is a benefit. I initially suggested this because there are multiple sets of function calls which are nearly identical, yet different, and could be combined into one call from one instance of the function.

> Perhaps you could re-name configuration variables such that they use `UPPER_SNAKE_CASE`?

Will do!

> We could just set `$err` to the error string and then `$embed.Set "description" $err` in our `if $err` block--is that consistent with what you're suggesting, or do you have a different approach in mind?

Similar to what I was suggesting, just with one less step. I was suggesting a *constant* `sdict` (`$errors`) with keys as error types and value as error output string. Each `if` block would set a variable (`$errType`) to an error type; then we would do `$errors.Get $errType` in the `if $err` block. Essentially the same thing, yours doesn't include the `sdict` as it isn't needed; I'll commit this soon.

> move the duplicate calls to `$embed.Set "color"` into a single line in the error handling block.

Will do!

H1nr1:
> I like this idea--as you said, it's only two arguments. I agree with `/heads?/` and `/tails?/`, but anything more just seems more complicated than it need be; this isn't Twitter, users need not worry about hitting any low character limit anytime soon and `h` / `t` just feels... unnecessary.

Less work for me, I'll get it in commit form soon!

H1nr1:
> Reading the diff you posted, I do like the approach you took. Using a map with a meaningful name and setting keys relating to the overall result as they come in. Your initial assignment of a side is also pretty neat, first time I actually have seen something like that, just the right amount of clever without being too obtuse.

Sounds like you're for it, I'll commit this soon too!

l-zeuch:
Reading through the code I do notice an issue with our current control flow: Encountering an error should **immediately** stop execution, lest we run into secondary errors (or worse, runtime errors).

For example, consider the user providing an invalid number that `toInt` cannot parse; `$bet` will now equal to 0, therefore failing the test for `lt $bet $MIN_BET`. This shadows our initial error (invalid number) with a different error (bet must be at least $MIN_BET), which is understandably quite confusing to a user. See attached screenshot for a demonstration.

image

To avoid deep nesting and keep readability to a reasonable level, we could consider using an associated template to handle errors, e.g.

```go
{{ define "err" }}
    {{ $err := .TemplateArgs.err }}
    {{ $embed := .TemplateArgs.embed }}
    {{ $embed.Set "description" $err }}
    {{ $embed.Set "color" 16488706 }}
    {{ $message := sendMessageRetID nil (cembed $embed) }}
    {{ deleteMessage .Channel.ID $message 5 }}
    {{ deleteTrigger }}
    {{ return }}
{{ end }}
```

Then call the template with `{{ return (execTemplate "err" (sdict "err" $err "embed" $embed)) }}`. This line surely reads a little weird, but this is due to the way `return` works--it goes one level of execution up, so we have to return `"err"`s return value once more to actually quit.

I do have a patch for this ready which I can push if need be, just let me know.

H1nr1:
Good spot! That went over our heads for too long.

I like your implementation of the template for error execution. However, I believe making the `if`s after the first error checking `if`, `else if`s would solve this problem without the need for a template. `$choice` and `$bet` would need to move to the top of the `if not .ExecData` block, while `$cd` and `$userBalance` can be defined in the condition to be used within their respective blocks.

```go
{{ $choice := reFind `(?i)(head|tail)s?` .StrippedMsg | lower }}
{{ $bet := reFind `\d+` .StrippedMsg | toInt }}

{{ if not (eq (len .CmdArgs) 2) }}
	{{ $err = printf "Syntax: **%scoinflip <heads/tails> <money>**" .ServerPrefix }}
{{ else if not $choice }}
	{{ $err = "You can only choose from **heads** or **tails**." }}
{{ else if not $bet }}
	{{ $err = printf "`%s` is not a valid number." (index .CmdArgs 1) }}
{{ else if lt $bet $MIN_BET }}
	{{ $err = printf "Minimum bet can only be **%d**!" $MIN_BET }}
{{ else if gt $bet $MAX_BET }}
	{{ $err = printf "Maximum bet can only be **%d**!" $MAX_BET }}
{{ else if $cd := dbGet .User.ID "cd_coinflip" }}
	{{ $err = print "You need to wait **" ($cd.ExpiresAt.Sub currentTime | humanizeDurationSeconds) "** to use this command again." }}
{{ else if lt ($userBalance := (dbGet .User.ID $DB_KEY).Value | toInt) $bet }}
	{{ $err = printf "You dont have **%d** credits.\nYou only have **%d** credits." $bet $userBalance }}
{{ else if dbGet .User.ID "isrolling_cf" }}
	{{ $err = "You can use this command when the last coin stopped rolling!" }}
{{ end }}
```

I've tested the above and it works as I intend, producing the first error it comes across, if any.

l-zeuch:
Either approach is fine by me--obviously I'm biased towards using a template (because that's how I implemented it), but if you decide to go the if-else chain route, that is fine as well.

For some pretending-to-be-objective commentary, if-else chains do tend to hurt readability, though it isn't much different to our current flow of checking the arguments.

src/fun/coinflip.go.tmpl
```diff
@@ -3,118 +3,105 @@
...
-			{{ $embed.Set "description" "You can only choose from **heads** or **tails**." }}
-			{{ $embed.Set "color" 16488706 }}
-			{{ $err = true }}
-		{{ end }}
-	{{ else }}
-		{{ $embed.Set "description" (printf "Syntax: **%scoinflip <heads/tails> <money>**" .ServerPrefix) }}
-		{{ $embed.Set "color" 16488706 }}
-		{{ $err = true }}
+	{{ if not (eq (len .CmdArgs) 2) }}
+		{{ $err = printf "Syntax: **%scoinflip <heads/tails> <money>**" .ServerPrefix }}
```
l-zeuch, line 25: Should we call this argument `bet` instead?

```suggestion
		{{ $err = printf "Syntax: **%scoinflip <heads/tails> <bet>**" .ServerPrefix }}
```

H1nr1:
I saw both having downsides; mine being readability, your template (as I'm told) is slower. I've never tested the difference in runtime of using a template call versus not, but I doubt it's enough to notice.

As the `if $err` block is essentially what would be the template call, I don't see an issue with keeping it as it is. That said, I think the `else if` route would be better and I can add line breaks between each if that helps spread out the code for better readability.

src/fun/coinflip.go.tmpl
```diff
@@ -3,118 +3,105 @@
...
+		{{ $err = print "You need to wait **" ($cd.ExpiresAt.Sub currentTime | humanizeDurationSeconds) "** to use this command again." }}
 	{{ end }}
+
+	{{ $userBalance := (dbGet .User.ID $DB_KEY).Value | toInt }}
+	{{ if lt $userBalance $bet }}
+		{{ $err = printf "You dont have **%d** credits.\nYou only have **%d** credits." $bet $userBalance }}
+	{{ end }}
+
+	{{ if dbGet .User.ID "isrolling_cf" }}
+		{{ $err = "You can use this command when the last coin stopped rolling!" }}
```
l-zeuch, line 57:
```suggestion
		{{ $err = "You can use this command once the last coin stopped rolling!" }}
```
I'm unsure which one is more correct; feel free to discard.

src/fun/coinflip.go.tmpl
```diff
@@ -3,118 +3,105 @@
...
 		{{ deleteMessage .Channel.ID $message 5 }}
 		{{ deleteTrigger }}
+		{{ return }}
 	{{ end }}
+
+	{{/* flip may proceed */}}
+	{{ dbSet .User.ID "isrolling_cf" true }}
+	{{ $embed.Set "image" (sdict "url" $img) }}
+	{{ $embed.Set "color" 16765696 }}
+	{{ $embed.Set "description" "**The coin is rolling... Wait...**" }}
```
l-zeuch, line 73: Politeness :-)

```suggestion
	{{ $embed.Set "description" "**The coin is rolling... Please wait...**" }}
```

l-zeuch (commented):
Random text-only commentary, the overall code looks fine (minus the current discussion re: error handling).

H1nr1:
I believe I will commit my `else if` method of error handling after thinking about users learning the YAGPDB CC language. I think it would be easier to comprehend due to `else if`s being very common in *all* programming languages.

H1nr1 (commented):
This review makes changes to the terms used for consistency and realism towards a coinflip.

l-zeuch:
Your self-review looks good! I left a comment about handling argument parsing; to summarise, we should re-evaluate whether it's worth the trade-off of having no positional arguments vs. having an easier time providing clear errors.

src/fun/coinflip.go.tmpl
```diff
@@ -3,118 +3,105 @@
...
+		{{ $err = printf "`%s` is not a valid number." (index .CmdArgs 1) }}
+	{{ end }}
+
+	{{ if lt $bet $MIN_BET }}
+		{{ $err = printf "Minimum bet can only be **%d**!" $MIN_BET }}
+	{{ end }}
+
+	{{ if gt $bet $MAX_BET }}
+		{{ $err = printf "Maximum bet can only be **%d**!" $MAX_BET }}
+	{{ end }}
```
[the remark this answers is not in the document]
H1nr1: I think the idea of having one error for no bet argument, then another for out of range bet argument would be enough to validate the bet argument while also providing clear errors. How I would go about this follows.
```go
{{ $bet := reFind `\d+` .StrippedMsg | toInt }}
{{ if not $bet }}
	{{ $err = printf "Please provide a bet between **%d** and **%d**" $MIN_BET $MAX_BET }}
{{ end }}
{{ if or (lt $bet $MIN_BET) (gt $bet $MAX_BET) }}
	{{ $err = printf "Bet must be minimum **%d** and maximum **%d**!" $MIN_BET $MAX_BET }}
{{ end }}
```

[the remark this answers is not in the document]
l-zeuch: How this is written it will always return "Bet must be minimum $MIN_BET and maximum $MAX_BET", because whilst `$bet = 0` would pass `if not $bet`, it is also less than $MIN_BET (which is at least 1 if it's supposed to make sense), therefore shadowing the initial assignment to `$err`.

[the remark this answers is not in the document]
H1nr1: > How this is written it will always return "Bet must be minimum $MIN_BET and maximum $MAX_BET", because whilst `$bet = 0` would pass `if not $bet`, it is also less than $MIN_BET (which is at least 1 if it's supposed to make sense), therefore shadowing the initial assignment to `$err`.

I forgot to use `else if`, which should solve this issue. I believe I copied the code from an earlier commit which used the if-if-if structure and made changes from there. If we were to change the bet validation to the above, it would have an if-else if structure.

src/fun/coinflip.go.tmpl
```diff
@@ -3,118 +3,97 @@
...
+		{{ $err = "You can only choose from **heads** or **tails**" }}
+
+	{{ else if not $bet }}
+		{{ $err = printf "Please provide a bet between **%d** and **%d**" $MIN_BET $MAX_BET }}
+
+	{{ else if lt $bet $MIN_BET }}
+		{{ $err = printf "Minimum bet can only be **%d**!" $MIN_BET }}
+
+	{{ else if gt $bet $MAX_BET }}
+		{{ $err = printf "Maximum bet can only be **%d**!" $MAX_BET }}
```
H1nr1, line 40:
```suggestion
{{ else if or (lt $bet $MIN_BET) (gt $bet $MAX_BET) }}
		{{ $err = printf "Bet must be minimum **%d** and maximum **%d**!" $MIN_BET $MAX_BET }}
```

H1nr1:
I believe the docs need some altering to match the state of the CC, as well as some general rewording. Should I start on that is this PR or open a new one?

l-zeuch:
Just include your documentation updates in this PR, they're logically connected changes.

H1nr1:
Feel free to suggest further edits or disagreements to my website changes.

I noticed the [slot machine](https://yagpdb-cc.github.io/fun/slot-machine) command has a caution message that states the command assumes the server has a currency system in place to be based on. Is this something we should add to the coinflip website too?

l-zeuch:
> Is this something we should add to the coinflip website too?

Yes, given that it also hooks into an already existing currency system, it would only make sense to also note that down.

H1nr1:
I thought so, but it was the only currency-based CC that I noticed include it; will add that now.

l-zeuch (approved):
Thanks for your work on this!

PR #354, opened 2025-02-11
Small change to reaction_handler.go.tmpl for better self-hosted compatibility

itzMiney:
Changed YAGPDB's User ID to .BotUser.ID for better compatibility with self hosted instances of YAGPDB

**Status**

### DIFF
--- a/src/leveling/reaction_handler.go.tmpl
+++ b/src/leveling/reaction_handler.go.tmpl
@@ -9,7 +9,7 @@
 {{ $validEmojis := cslice "▶️" "◀️" }} {{/* Valid emojis */}}
 {{ $isValid := false }} {{/* Whether this is actually a valid embed / leaderboard embed */}}
 {{ $page := 0 }} {{/* The current page */}}
-{{ with and (eq .ReactionMessage.Author.ID 204255221017214977) .ReactionMessage.Embeds }} {{/* Checks for validity */}}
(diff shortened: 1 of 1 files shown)


H1nr1:
Assuming this idea is accepted, there are many other instances that could be changed, excluding command triggers which do not have access to `.BotUser`.

galen8183:
> Assuming this idea is accepted, there are many other instances that could be changed, excluding command triggers which do not have access to `.BotUser`.

It would be preferable to always set `.BotUser` regardless of member state, since it is (relatively) constant.

itzMiney:
I did see that issue on a couple other commands too yes, .BotUser is definitely more flexible than just hard coding it to use yag's id

l-zeuch:
Thanks for the PR. If it's not too much trouble, would you be willing to expand the scope of this patch to touch up all the custom commands that reference the bot's ID in some way, preferring `.BotUser` and its fields?

itzMiney:
Sure, shouldn't be too hard

itzMiney:
I think that should be all ccs that use yags id in them, for trigger people that self host it hopefully have enough brains to figure out that it is yags id and they need to change it to their own bot's id

l-zeuch:
Thanks for your work on this!

savage4618:
as reported here: https://discord.com/channels/166207328570441728/578976698931085333/1341561752030154813
This PR introduced a bug in the editmessage CC

itzMiney:
Oops my bad just woke up and saw this, must've overlooked that one instance being in a control structure
I see you already made a quick PR that fixes that small oversight 👍

PR #365, opened 2025-05-16
Update utilities/snipe

H1nr1:
__General update__
- Change undelete regex: capture first (most recent) message log *entirely*
- Update guild icon generation method
- Reorder and change statements for embed building so embed updates are included on send

**Status**

### DIFF
--- a/src/utilities/snipe.go.tmpl
+++ b/src/utilities/snipe.go.tmpl
@@ -31,7 +31,7 @@
 {{ $msg := index $data 3 }}
 
 {{ if $REDACT_INVITES }}
-	{{ $msg = reReplace `(?:https?:\/\/)?(?:www\.)?(?:dsc\.gg|invite\.gg+|discord\.link)\/[\w\d-_]+|(?:https?:\/\/)?(?:www\.|ptb\.|canary\.)?(?:discord(?:app)?\.(?:(?:com|gg)\/invite\/[\w\d-_]+)|discord\.gg\/[\w\d-_]+)|(?:https?:\/\/)?(?:www\.|ptb\.|canary\.)?(?:discord(?:app)?\.(?:(?:com|gg)\/(?:invite|servers)\/+[\w-_]+)|discord\.gg\/+[\w-_]+)|(?:https?:\/\/)?(?:www\.)?(?:dsc\.gg|invite\.gg+|discord\.link|(?:discord\.(gg|io|me|li|id))|disboard\.org)\/[\w\d-_/]+` $msg "<*redacted invite link*>" }}
+	{{ $msg = reReplace `(?:https?://)?(?:www\.|ptb\.|canary\.)?(?i:discord(?:app)?|dsc|invite|disboard)\.(?i:gg|com|io|me|li|id|org)/\S+` $msg "<*redacted invite link*>" }}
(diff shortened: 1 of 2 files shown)


H1nr1:
In the midst of writing this comment, I have realized the "remove invite link" idea was not implemented in the previous code. I got it from a member of the YAGPDB Support server which included it in their code. This is the fault of me copying their code to test, and going further using it.

I am wondering whether their version of the snipe code is dated or if this was a custom addition somewhere along the way. Either way, whether to keep it or rid it is on the table.

I will include my initial response intended for this comment below.

I would like to simplify the invite link regex, if possible

The issue being I am not confident enough in the specifics of each link type, such as whether

`(?:https?:\/\/)?(?:www\.|ptb\.|canary\.)?(?:discord(?:app)?\.(?:(?:com|gg)\/invite\/[\w\d-_]+)` and 

`(?:https?:\/\/)?(?:www\.|ptb\.|canary\.)?(?:discord(?:app)?\.(?:(?:com|gg)\/(?:invite|servers)\/+[\w-_]+)`

can be combined to the following (notice last segment)

`(?:https?:\/\/)?(?:www\.|ptb\.|canary\.)?(?:discord(?:app)?\.(?:(?:com|gg)\/(?:invite|servers)\/+[\w\d-]+)`

galen8183:
> whether to keep it or rid it is on the table.

This would be a good config option, disabled by default.

> simplify the invite link regex

`\w` is `[A-z0-9_]`, so `[\w\d]` is redundant.

H1nr1:
I thought about making it a configurable option, but didn't want to complicate things too much. Will make it configurable now though.

As for the regex, thank you, I was too focused on removing the `_` as `\w` covers that as well.

What do you think about combining the ORs into one main domain, if you will, and not worrying about the specifics of what each subdomain and whatnot allows to be a valid invite? There could end up with some false positives, but even if they aren't technically valid, they were likely an attempt to advertise. I hope that makes sense.

galen8183:
> didn't want to complicate things too much

It could be as small as making the `reReplace` replacement string a config opt with a default value of `"$0"`, and have `reReplace` run unconditionally

> What do you think about combining the ORs into one main domain

Agreed, something like this might work well enough: `(?:https?://)?(?:www\.|ptb\.|canary\.)?(?i:discord(?:app)?|dsc|invite|disboard)\.(?i:gg|com|io|me|li|id|org)/\S+`

H1nr1:
I think a simple boolean configurable variable would be best for the user. If they wanted more, the replacement string should be easy enough to change. Let me know if that is not acceptable though.

Your first regex example was what I was pushing for; I'll use it for now. Thanks!

H1nr1:
It was brought up by the member that initially motivated me to start this PR that they wanted to be able to provide a numerical argument for which message to view. Such as `-snipe 3` would present the third most recently deleted message, instead of the most recent every time.

Would it be worthwhile to add this as a feature, or is that encroaching on the 'undelete' command too much?

H1nr1:
I believe this is up to date and ready to be merged if you are good with it!

l-zeuch:
I will review this on the weekend, thanks in advance for your work on this!

l-zeuch (approved):

PR #368, opened 2025-08-30
fix(qotd): add $allPerms sdict to resolve undefined variable error

### LINKED ISSUE #367: Starboard V2
When I'm reacting a msg or pic it sending the error but it also working and sending the starboard in the specified channel 
**The Error**

`An error caused the execution of the custom command template to stop:
`Failed executing CC #4, line 39, row 81: executing "CC #4" at <$thisID>: nil pointer evaluating *discordgo.Message.Author`
``38    
39        {{if and .ReactionAdded (n...
40            {{deleteMessageReactio...

- ``

Image

hazelhop:
This PR fixes the QOTD setup custom command by defining the missing $allPerms sdict.
Without this, the command fails with "undefined variable $allPerms".

Changes:
- Added explicit $allPerms permission mapping.
- Updated MissingPerms template call.

Tested and working as expected.

Issue was in component handler cc

### DIFF
--- a/src/fun/qotd/advanced/component_handler.go.tmpl
+++ b/src/fun/qotd/advanced/component_handler.go.tmpl
@@ -2,9 +2,9 @@
   {{ $yPerms := getTargetPermissionsIn .Bot .CID }}
   {{ $missing := cslice }}
   {{ range $k, $v := .AllPerms }}
-    {{ if and (in $.ReqPerms $v) (ne (bitwiseAnd $yPerms $v) $v) }}
-      {{ $missing = $missing.Append $k }}
-    {{ end }}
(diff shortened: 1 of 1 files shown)


jo3-l (changes requested):
Thanks for the PR. I'm not convinced the definition of `$allPerms` is the issue. The original code already defines `$allPerms := .Permissions`. I think the real fix you made is passing `"Bot" $.BotUser.ID` to each invocation of the `MissingPerms` template.

Could you check that just fixing the `execTemplate` calls suffices? If not, could you briefly explain what the problem is with the current definition of `$allPerms`?

If the above is too much trouble, myself or one of the other maintainers can check this out ourselves when time permits. Thanks again.

Mirivel:
Excelente

Mirivel (approved):

PR #388, opened 2026-04-16
Add export-ccs CC

galen8183:
- **feat: export CCs CC**
- **feat(cc2file): mention new export-ccs CC**

There has been (especially recent) demand for exporting CCs, hopefully this script will give people enough utility and be easier for VS to point to :-)

**Status**

### DIFF
--- a/src/utilities/export-ccs.go.tmpl
+++ b/src/utilities/export-ccs.go.tmpl
@@ -6,15 +6,11 @@
 */}}
 
 {{/* Configuration values */}}
-{{$permissions := cslice .Permissions.Administrator .Permissions.ManageServer}}
+{{$permissions := bitwiseOr .Permissions.Administrator .Permissions.ManageServer}}
 {{/* End of configuration values */}}
 
 {{if not .ExecData}}
-	{{$hasPerms := false}}
-	{{range $permissions}}
-		{{- $hasPerms = or $hasPerms (hasPermissions .) -}}
-	{{end}}
-	{{if not $hasPerms}}
+    {{if not (hasPermissions $permissions)}}
 		You don't have permission to run this command!
 		{{return}}
 	{{end}}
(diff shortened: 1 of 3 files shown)


src/utilities/export-ccs.go.tmpl
```diff
@@ -0,0 +1,122 @@
+{{/*
+	Export all server CCs as files.
+	See <https://yagpdb-cc.github.io/utilities/export-ccs> for more information.
+
+	Author: galen8183 <https://github.com/galen8183>
+*/}}
+
+{{if not .ExecData}}
+	{{if not (hasPermissions .Permissions.Administrator)}}
```
l-zeuch: Do we want to extend this to, say, `ManageServer`? Or maybe even as a configuration value with a sane default? Though, given that it is fairly high up, users should be able to easily spot it and edit accordingly.

website/docs/utilities/export-ccs.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Export CCs
+---
+
+Export all custom commands in the server as files with additional metadata.
+
+:::note
```
l-zeuch: I think a tip admonition is suited better here

```suggestion
:::tip
```

l-zeuch (approved):
Thanks! I have just two remarks, noted below.

src/utilities/export-ccs.go.tmpl
galen8183: I'm mostly indifferent on the specific perms, ManageServer makes enough sense and a config variable is simple enough that adding one sounds fine to me :-)

l-zeuch:
I have refactored your permission detection to use bitwise operations instead, saving us the need to range over however many permissions (though probably not a whole lot of them).

It is in my opinion also a little more idiomatic, so users can learn how to use bitops to their advantage.

galen8183:
Good point, thanks!

PR #390, opened 2026-05-03
fix(birthday): precise time calc & db cleanup logic

EsenDR:

1. The current age verification check ignores leap years and relies on a harcoded hour count. This PR replaces some time calculations in the Birthday CC to rely on Go's native `.AddDate` method instead of hardcoded integer math.
2. Current script attempts to clean up yesterday's `bdayannounced` DB flag using `mult -24 $.TimeHour | toDuration`. But `toDuration` interprets ints as nanoseconds causing it to always evaluate to the current day and failing to delete the previous day's flag. This PR replaces that with `currentTime.AddDate 0 0 -1` to guarantee getting the previous day.
3. `-bday start` check used similar nanosecond math to verify that users couldn't schedule the loop past tomorrow. This PR simplifies it to calculate `$target.Day` against `$tomorrow.Day` via `.AddDate 0 0 1`.
4. Currently, if a user runs `-bday del` and they were the only person for that day, it left an empty array `[]` taking up space in the DB. This PR updates `handleDeletes` to check list lengths. If the day's array drops to 0, the day key is deleted. If the month `sdict` drops to 0, the entire month key is deleted from the DB.
5. Bits of syntax cleanup to make the code feel more readable for `$error`s (this is subjective, please let me know if it doesn't work).

**Note:**

I would like more testers to test this if possible. I am just one person, and I am not great with scripting & testing. I was able to test my changes and confirm my changes work, but any more testing would be appreciated. The Database page for the CCs is a great tool that assisted me with testing these changes.

**Status**

### DIFF
--- a/src/birthday/birthday.go.tmpl
+++ b/src/birthday/birthday.go.tmpl
@@ -1,6 +1,6 @@
 {{/* User Variables */}}
 {{$mods := cslice 748955521255473202}}
-{{$channelID := 741395967638634499}}
+{{$channelID := 741395967638634499}} {{/* Channel ID to send the bday msgs */}}
 {{$bdayMsg := "Congratulations for your birthday!"}}
 {{$invertedOrder := true}}
 {{$yearOptional := true}}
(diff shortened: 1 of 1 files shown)


l-zeuch (approved):
LGTM. Not a fan of these levels of indentation, more specifically the `with`-spam, but we can clean that up in later modernisation passes.

